PR #23, opened 2015-03-16
Added ability to pass flags to mysqld process

CedricGatay:
Typically one can need to pass flags when spawning a new MariaDB/Mysql process like `lower_case_table_names` for example when using mixed environments.
This is now possible by using the `addMysqldArgs`  method in the `DBConfigurationBuilder`.

### DIFF
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
@@ -163,6 +163,9 @@ public class DB {
         if (!config.isWindows()) {
             builder.addArgument("--socket=" + getAbsoluteSocketPath());
         }
+        for (String arg : config.getMysqldArgs()) {
+            builder.addArgument(arg);
+        }
         cleanupOnExit();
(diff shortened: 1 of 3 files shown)


kaismh:
This commit also solves the need for hardcoding UTF-8, since it can be passed as an argument

vorburger:
This is a great idea! Just merged as 8da9aa631409a7a2416cf892c4f08fc3675c05be. Please note my renaming in 41912c9daf06678884d202b8daada468b7edc01c. Thanks a lot for contributing to MariaDB4j, please keep more like this coming! ;-)

CedricGatay:
Glad to help and contribute to a great project!

krm1312:
As a workaround on 2.1.3 version (since that is latest released artifact), you can do something like the following for dynamic variables at least.

DB db = DB.newEmbeddedDB(mariaDBConfiguration());
        db.start();
        db.run("set global innodb_file_per_table=true");
        db.run("set global innodb_file_format=Barracuda");
        db.run("set global innodb_large_prefix=true");
        db.run("set global character_set_server=utf8mb4");
        db.run("set global collation_server=utf8mb4_unicode_ci");

...do what you need here...

PR #31, opened 2015-11-07
Synchronize db install.

krm1312:
Syncrhonize db install to try to fix some intermittent test
failures when running parallel tests in maven that depend
on MariaDB4j.


vorburger:
@krm1312 sure, tx! PS: Sorry it took so (too) long to accept & merge this.. life.

PR #32, opened 2015-11-10
Upgrade Support for MariaDB 10.0.21 for Linux

HaVonTe1:
regarding to the issue: Missing Support for MariaDB 10.x for Linux
Sorry for the late PR.


vorburger:
@HaVonTe1 Thank You for this contribution, and so sorry I'm only picking this up now!

I just noticed that @	timorohwedder has meanwhile, just a week after you had raised this, put in https://github.com/vorburger/MariaDB4j/pull/33, and as that is with a more recent 10.1.8 (yours here was 10.0.21), I hope you don't mind that I close this and integrate that other one instead.

PR #38, opened 2016-06-17
update linux version to 10.1.13

ghiron:


vorburger:
merged as 22d5d664f5d3293ec7d337d94363274f0a3aa3d3

@ghiron  thanks for this contribution! FYI a deploy of a new 2.2.1 including this to Maven Central should follow within the hour or so, hopefully.

Hope that it is OK that I've also added you to the new https://github.com/vorburger/MariaDB4j/blob/master/USERS.md & https://github.com/vorburger/MariaDB4j/blob/master/CONTRIBUTORS.md files in 4aa6fc44cc5bd12e4cfcdd70489c528d33972953 (FYI that's now included in the mariaDB4j JAR via 243471891ab6bdb61fea1020c85d6ecec23ec300).

Also note the new section on the README re. DB upgrades just introduced in 4c2da73569a59696db92c0963a7d21964cc3401f. I haven't done this yet for this contribution just because I'm lazy, but I think that's how we should do it in the future. Please shout if you think that this is stupid for some reason that I'm missing.

ghiron:
:+1:

PR #40, opened 2016-07-18
Made libs dir relative to base dir unless it is explicitly set.

anverus:


vorburger:
merged as 91ea1e287114bd322439c8b71635d5a8ceb2506c

PR #44, opened 2016-07-26
Issue37 add mac maria db10x support

hanklank:
Runs fine when tested locally with our integrationtests.


hanklank:
Note: ci is failing because the 10.1.9 isn't in repo yet as it never been built, and has to be built first for the dependencies to find it.

vorburger:
merged (squashed) to master as 286dfffae8bd1278ef44d6c8856f7166d10c469e

Issue #36, opened 2016-04-06
Don't delete baseDir when it is set to /tmp/


simonft:
Currently [cleanupOnExit](https://github.com/vorburger/MariaDB4j/blob/9848055af3a10484134c3f3d6b81ed17d14dd77f/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java#L355) deletes the baseDir if it thinks it is a temporary directory, which is determined with [isTemporaryDirectory](https://github.com/vorburger/MariaDB4j/blob/9848055af3a10484134c3f3d6b81ed17d14dd77f/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/Util.java#L89).

isTemoraryDirectory uses directory.startsWith(SystemUtils.JAVA_IO_TMPDIR) to determine whether it is a "temporary directory". I think there might be two different definitions of temporary directory used here though. cleanupOnExit should delete the baseDir if it is a directory which is temporary, as in around only for a short period of time. SystemUtils.JAVA_IO_TMPDIR will tell you if it is a directory for storing temporary files, not if it's a directory that can be removed without issue. The best example of this is '/tmp/', which is what we had it set to.

In fact, cleanupOnExit should not delete any directories it has not created itself.

vorburger:
I see what you mean. Would you be willing to contribute to the project and
put in a pull request with a suggested change? I'd definitely look at it
for possible merge.

You can also subclass it and override isTemporaryDirectory - I've
intentionally made it a protected for this.. ;-)

vorburger:
@simonft if you would like to contribute a PR with what you had in mind above, I'd certainly look at it... I just tried a quick implementation myself, but "should not delete any directories it has not created itself", which does sound good to me, is not 1 line trivial change to implement, what would be the exact criteria for you for "it has not created itself" ? For the baseDir, would it be if a call through installPreparation() ? Or unpackEmbeddedDb() ? For dataDir, would it be .. what, createDB() ? Nah, that's not reliable / what you want. Hm.

paulroemer:
What about letting the user decide if he/she wants to delete the directory during shutdown. Wouldn't it be much more easier to just add an option to the configuration?

```configBuilder.setKeepDataDir()```

vorburger:
@paulroemer that would work for me .. for backwards compat. perhaps the default should be to behave as it's now, but users of a new version could change it via the setter on configuration?

vorburger:
[...]

PR #60, opened 2017-02-10
#36: Adds a new configuration option to prevent the deletion of temporary directories

paulroemer:
A first try and base for discussion.

### DIFF
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfigurationBuilder.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfigurationBuilder.java
@@ -48,6 +48,7 @@ public class DBConfigurationBuilder {
     protected String dataDir = DEFAULT_DATA_DIR;
     protected String socket = null; // see _getSocket()
     protected int port = 0;
+    protected boolean deletesTemporaryDataOnShutdown = true;
     protected boolean isUnpackingFromClasspath = true;
     protected List<String> args = new ArrayList<String>();
 
@@ -113,6 +114,22 @@ public class DBConfigurationBuilder {
         return this;
     }
 
+    public boolean deletesTemporaryDataOnShutdown() {
+        return deletesTemporaryDataOnShutdown;
+    }
+
+    /**
+     * Defines if the configured data and base directories should be deleted on shutdown.
+     *
+     * @param deletesTemporaryDataOnShutdown
+     * @return this
+     */
+    public DBConfigurationBuilder setDeletesTemporaryDataOnShutdown(boolean deletesTemporaryDataOnShutdown) {
+        checkIfFrozen("keepsDataAndBaseDir");
+        this.deletesTemporaryDataOnShutdown = deletesTemporaryDataOnShutdown;
+        return this;
+    }
+
     protected int detectFreePort() {
         try {
             ServerSocket ss = new ServerSocket(0);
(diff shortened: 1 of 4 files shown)


paulroemer:
Hi Michael,

here my first suggestion. For now I just added some simple stupid tests that check if setting the parameter works.

What is discussable:
1) The name of the parameter
2) Am i allowed to refactor the shutdown handler to make testing possible and to add a mocking framework you like to test it?

Cheers,
  Paul

vorburger:
@paulroemer thanks for taking this on! (And sorry for delayed reply.) This would work for me...

Re. 1. naming, how about deletesTemporaryFilesOnShutdown instead of deletesTemporaryDataOnShutdown? Or even deletesTemporaryBaseAndDataOnShutdown? Just because deletesTemporaryDataOnShutdown sounds a bit like its affecting the data dir only, when in fact it's for both (which is fine with me). Change it everywhere, if OK with you.

Re. 2. "allowed to refactor the shutdown handler" - of course! As long as everything still works, you can contribute whatever you think can be improved - most welcome. (Keep public API comparability in mind, for the main functionality; for the shutdown handler it shouldn't matter.)

Re. mocking framework, if you must use one, my preference is Mockito - BUT, personally, and I do realise this is not commonly accepted... ;-) I'm, in general, not a huge fan of funky mocking frameworks.. I've seen too many tests of people mocking too much. I just write simple unit tests where I often don't find the need for mocking frameworks, and do like to write integration tests, e.g. db.stop() and assert dir still there IFF deletesTemporaryFilesOnShutdown false...

... I see you can't easily do that, but ... in looking over my code again, I'm not sure anymore if at the time I had had a particular reason for doing this FileUtils.deleteQuietly() only in the ShutdownHook - if I did this again, I would seem more natural to me now to just always do this in the db.stop() instead? What do you think? If you like this, feel free to put in a change to move that - perhaps best in a separate PR first?

vorburger:
@paulroemer if you have a chance to finish that up, I'd love to merge it! No stress, but FYI I'm probably going to push a new release "soon" with what's coming from https://github.com/vorburger/MariaDB4j/pull/71.

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfigurationBuilder.java
```diff
@@ -113,6 +114,22 @@ public DBConfigurationBuilder setPort(int port) {
         return this;
     }
 
+    public boolean deletesTemporaryDataOnShutdown() {
```
vorburger: is...

vorburger:
@paulroemer never got to this, so I'll see if I can finish this up, based on the proposals made above.

vorburger:
Closing, as I've finished this off in https://github.com/vorburger/MariaDB4j/pull/84. Will do an IT separately in next.

@paulroemer I only realized looking over this now that you were @ vaadin.com ... :) I've used Vaadin in the past - good stuff! Cool to see you guys use my MariaDB4j presumably in one of your customer projects? Feel free to raise a PR to add you / Vaadin to the [USERS](https://github.com/vorburger/MariaDB4j/blob/master/USERS.md) and/or [CONTRIBUTORS](https://github.com/vorburger/MariaDB4j/blob/master/CONTRIBUTORS.md) files.

[Standard Footer] PS: Shameless plug, consider https://www.patreon.com/vorburger to support MariaDB4j.

Issue #66, opened 2017-04-18
Password prompt when using dbInstance.source(..)


bjornblomqvist:
When using dbInstance.source(..) method with an empty string as password a prompt is shown asking for the password.

    db.source("sql/schema.sql", "root", "", "test_db");

I think the method should either validate input and throw an error when an empty string is supplied or treat it the same as a null value.

vorburger:
@bjornblomqvist thanks for your interest in this project. A Pull Request to remedy this would be welcome!

bjornblomqvist:
So what solution do you think is the right way to go? Also do you need a test for this before you would accept a pull requests?

vorburger:
> So what solution do you think is the right way to go? 

I like your idea of just treating an emptyString the same as a null value, probably by just using `password != null && !password.isEmpty()` in https://github.com/vorburger/MariaDB4j/blob/master/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java#L259 ... you could also do it not just for password but also for username & dbName perhaps?

> Also do you need a test for this before you would accept a pull requests?

Yes that would be great, please. I'm assuming that it should be trivial to create one... just add an additional line to https://github.com/vorburger/MariaDB4j/blob/master/mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleTutorialTest.java#L75 with "" instead of null - as that tests sources() with null already.

vorburger:
Closing as Done, following merge of https://github.com/vorburger/MariaDB4j/pull/67!

PR #67, opened 2017-04-19
Changed so that empty password string is treated the same as null

bjornblomqvist:
Previously the empty password string was passed on to the mysql
command line tool resulting in mysql asking for a password and
stalling the test run.

Solves issue #66


vorburger:
@bjornblomqvist thank you very much for your contribution to MariaDB4j!

Issue #65, opened 2017-04-07
Include under binaries the backup/restore tools


cortiz:
mysqldump and mysqlimport should be available so we could make portable backups.

vorburger:
@cortiz thanks for your interest in this project. I've nothing against including additional such binaries into the DB JARs, and perhaps even having methods in the Java APIs to nicely expose these utilities?

If you would like to contribute this to the project, please [see the relevant section on the README](https://github.com/vorburger/MariaDB4j/blob/master/README.md#mariadb-database-jars-and-version-upgrades).

cortiz:
This should be solve in PR#71

vorburger:
Closing this as a duplicate of the now ongoing and (if you could incorporate the review feedback) about to be merged https://github.com/vorburger/MariaDB4j/pull/71 - no need t keep this issue here open separately anymore now.

PR #71, opened 2017-05-19
Adds Export/dump API  #65

cortiz:
Since it's a new Binary and existing version were already publish an also upgrade all maria db version to be the same (10.1.23) I did the tests on Linux (all test passed) and did the same on a MacOS and Windows VMs and same as Linux all Junit test passed.

I also add a `MariaDB4jSampleDumpTest.java` to test and put and example on how to use export/dump feature.

To Restore a DB using `db.source` works


mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+
+    public void checkDBIntegrity(){
+        Connection conn;
+        try {
+            conn = DriverManager.getConnection(config.getURL(DBNAME), "root", "");
+            QueryRunner qr = new QueryRunner();
+            // Should be able to create a new table
+            List<String> resutls=qr.query(conn, "SELECT * FROM crew;",new ColumnListHandler<String>());
```
vorburger: typo, change to "results" instead of "resutls"

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * @author Michael Vorburger
+ * @author Michael Seaton
+ * @author Carlos Ortiz
+ */
+public class MariaDB4jSampleDumpTest {
+
+    private  DB db;
+    private  DBConfigurationBuilder config;
+    public  final String DBNAME="planetexpress";
```
vorburger: This looks like it's a (private) constant, so perhaps you'd like to make this static, and private, and with correct spaces:

    private static final String DBNAME = "planetexpress";

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Tests the functioning of MariaDB4j Dump/Restore illustrating how to use MariaDB4j.
+ *
+ * @author Michael Vorburger
+ * @author Michael Seaton
```
vorburger: you can remove my and @mseaton names here - **YOU** wrote this test - and should let the world know about it! ;-)

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
```diff
@@ -376,4 +376,55 @@ public void run() {
         });
     }
 
+
+    synchronized public ManagedProcess xmlDumpDB(File outputFile, String dbName,String user,String password)
+            throws IOException, ManagedProcessException {
+        return dumpDB(outputFile, Arrays.asList(dbName),true,true,true, user, password);
+    }
+
+    synchronized public ManagedProcess dumpDB(File outputFile, String dbName,String user,String password)
```
vorburger: let's name this method `dumpSQL`, because from what I understand, from your test, this is doing a SQL dump, as opposed to the other method which is doing a XML dump, so let the method names reflect that?

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
```diff
@@ -376,4 +376,55 @@ public void run() {
...
+            throws IOException, ManagedProcessException {
+      return dumpDB(outputFile, Arrays.asList(dbName),true,true,false, user, password);
+    }
+
+    synchronized public ManagedProcess dumpDB(File outputFile, List<String> dbNamesToDump,
```
vorburger: so name this one just `dump` without *DB suffix; and just to have a cleaner public API (and make it e.g. easier for users to choose on Ctrl-Space completion in an IDE), how about keeping this method here protected? Not private, but protected - if someone really wants, she can extend the DB class to access it - but we'd expect that in normal use cases, the xmlDump() or sqlDump() is what one would typically use.

vorburger: I'm not sure you want to make this method "synchronized" ... especially because it has a boolean lockTables - in theory, it would be OK to call this one concurrently, wouldn't it, to (in theory) run two dumps in parallel.. or is there something I'm missing?

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
```diff
@@ -376,4 +376,55 @@ public void run() {
         });
     }
 
+
+    synchronized public ManagedProcess xmlDumpDB(File outputFile, String dbName,String user,String password)
```
vorburger: would you mind to re-name this method `dumpXML()` intead?  I'd find that would express the "action" of this method better. (And there is no need for the *DB suffix, as this entire class is DB.)

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
```diff
@@ -376,4 +376,55 @@ public void run() {
...
+      return dumpDB(outputFile, Arrays.asList(dbName),true,true,false, user, password);
+    }
+
+    synchronized public ManagedProcess dumpDB(File outputFile, List<String> dbNamesToDump,
+                                              boolean compactDump, boolean lockTables, boolean asXml,
+                                              String user, String password)
+            throws ManagedProcessException, IOException {
+
+        ManagedProcessBuilder builder = new ManagedProcessBuilder(newExecutableFile("bin","mysqldump"));
+        /** Not used **/
```
vorburger: what is not used?

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
```diff
@@ -376,4 +376,55 @@ public void run() {
...
+        builder.addStdOut(new BufferedOutputStream(new FileOutputStream(outputFile)));
+        builder.setOutputStreamLogDispatcher(new OutputStreamLogDispatcher(){
```
vorburger: let's re-use `getOutputStreamLogDispatcher()` & `MariaDBOutputStreamLogDispatcher` here... I suspect that you saw that, but want the STDOUT to be logged as TRACE instead of as INFO, to avoid spamming your logs? But the way you're doing it, you're not handling type == OutputStreamType.STDERR, so it would loose the ability to log error messages from the dump, which seems kind of important to have in the log, agreed? So I think the best would be to reuse the existing `MariaDBOutputStreamLogDispatcher` - but if you need to, you can, of course, extend it to have a new constructor with something like:

    MariaDBOutputStreamLogDispatcher() {
        this(Level.INFO);
    }

    MariaDBOutputStreamLogDispatcher(Level stdoutLogLevel) { ...

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+ * @author Carlos Ortiz
+ */
+public class MariaDB4jSampleDumpTest {
+
+    private  DB db;
+    private  DBConfigurationBuilder config;
+    public  final String DBNAME="planetexpress";
+
+
+    public void beforeTest(){
```
vorburger: use JUnit's `@Before` annotation here

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+            e.printStackTrace();
+            fail(e.getMessage());
+        } catch (SAXException e) {
+            e.printStackTrace();
+            fail(e.getMessage());
+        }
+        afterTest();
+    }
+
+    public  void afterTest(){
```
vorburger: use JUnit's `@After` annotation here

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+            ex.printStackTrace();
+            fail(ex.getMessage());
+        } catch (ParserConfigurationException e) {
+            e.printStackTrace();
+            fail(e.getMessage());
+        } catch (SAXException e) {
+            e.printStackTrace();
+            fail(e.getMessage());
+        }
+        afterTest();
```
vorburger: remove this - JUnit will automatically call your `@After`

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+            assertTrue(FileUtils.sizeOf(outputDumpFile)>0);
+            FileUtils.forceDeleteOnExit(outputDumpFile);
+        }catch (ManagedProcessException ex){
+            ex.printStackTrace();
+            fail(ex.getMessage());
+        }catch (IOException ex){
+            ex.printStackTrace();
+            fail(ex.getMessage());
+        }
+        afterTest();
```
vorburger: remove this - JUnit will automatically call your `@After`

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+        }catch (Exception ex){
+            ex.printStackTrace();
+            fail(ex.getMessage());
+        }
+    }
+
+    @Test
+    public void simpleDump(){
+        try {
+            beforeTest();
```
vorburger: remove this - JUnit will automatically call your `@Before`

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+            fail(ex.getMessage());
+        }
+        afterTest();
+    }
+
+
+    @Test
+    public void XMLDump(){
+        try {
+            beforeTest();
```
vorburger: remove this - JUnit will automatically call your `@Before`

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+            fail(e.getMessage());
+        } catch (SAXException e) {
+            e.printStackTrace();
+            fail(e.getMessage());
+        }
+        afterTest();
+    }
+
+    public  void afterTest(){
+        try {
```
vorburger: remove, as above

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+            db.start();
+            db.createDB("planetexpress");
+            db.source("ch/vorburger/mariadb4j/dumpTest.sql");
+        } catch (ManagedProcessException e) {
+            e.printStackTrace();
+            fail(e.getMessage());
+        }
+    }
+
+    public void checkDBIntegrity(){
```
vorburger: why not just do this inline right at the end of & part of `beforeTest()` ? It makes the test easier to read.

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+    private  DB db;
+    private  DBConfigurationBuilder config;
+    public  final String DBNAME="planetexpress";
+
+
+    public void beforeTest(){
+        config = DBConfigurationBuilder.newBuilder();
+        config.setPort(0);
+        try {
```
vorburger: don't do this pattern (`try` / `catch` / `log` / `fail`) in a `@Test` - simply remove all this, and just let it `throws ManagedProcessException` - and JUnit will do the right thing for you!

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+            ManagedProcess dumpProcess = db.xmlDumpDB(outputDumpFile,DBNAME,"root","");
+            dumpProcess.start();
+            assertTrue(outputDumpFile.exists() || outputDumpFile.isDirectory());
+            assertTrue(FileUtils.sizeOf(outputDumpFile)>0);
+            // We just want to check that the file is a valid xml, output of it it's mysqldump resolvability .
+            DocumentBuilderFactory dbFactory = DocumentBuilderFactory.newInstance();
+            DocumentBuilder dBuilder = dbFactory.newDocumentBuilder();
+            Document doc = dBuilder.parse(outputDumpFile);
+            FileUtils.forceDeleteOnExit(outputDumpFile);
+        }catch (ManagedProcessException ex){
```
vorburger: remove all this, as above (and [use a multi catch](http://docs.oracle.com/javase/7/docs/technotes/guides/language/catch-multiple.html), next time you have to do something like this somewhere else; but you don't need it here)

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+        try {
+            beforeTest();
+            checkDBIntegrity();
+            File outputDumpFile = File.createTempFile("sqlDump ",".sql");
+            ManagedProcess dumpProcess = db.dumpDB(outputDumpFile, DBNAME,"root","");
+            dumpProcess.start();
+            assertTrue(outputDumpFile.exists() || outputDumpFile.isDirectory());
+            assertTrue(FileUtils.sizeOf(outputDumpFile)>0);
+            FileUtils.forceDeleteOnExit(outputDumpFile);
+        }catch (ManagedProcessException ex){
```
vorburger: remove this, as explained above

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+    public void checkDBIntegrity(){
+        Connection conn;
+        try {
+            conn = DriverManager.getConnection(config.getURL(DBNAME), "root", "");
+            QueryRunner qr = new QueryRunner();
+            // Should be able to create a new table
+            List<String> resutls=qr.query(conn, "SELECT * FROM crew;",new ColumnListHandler<String>());
+            assertEquals(4,resutls.size());
+            assertEquals("John A Zoidberg",resutls.get(2));
+        }catch (Exception ex){
```
vorburger: remove this, as explained above

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java
```diff
@@ -0,0 +1,155 @@
...
+    public void beforeTest(){
+        config = DBConfigurationBuilder.newBuilder();
+        config.setPort(0);
+        try {
+            db = DB.newEmbeddedDB(// 0 => autom. detect free port
+                    config.build());
+            db.start();
+            db.createDB("planetexpress");
+            db.source("ch/vorburger/mariadb4j/dumpTest.sql");
+        } catch (ManagedProcessException e) {
```
vorburger: remove also

cortiz:
@vorburger I did all the changes request + I also change `ch.vorburger.exec.ManagedProcessTest#testMustTerminateExec` to be `sleep` instead of `vi` the reasoning of this is that macOS vs Linux vi can be different between them and even between Linux distro, I went for a more standard  `sleep` that should get the job done. For example in my OSX VM and Arch every-day machine the test always fails because `vi` never start.

vorburger:
@cortiz thanks! I saw that you misunderstood the comment re. the xmlDumpDB method name (you mmade it xmlDump instead of my suggsted dumpXML), but never - I'm just adjusting that (and a few minor other things), and will then merge this - as a single squash merge, hope OK for you...

vorburger:
@cortiz FYI I've done some Git rebase interactive to squash your commits around a bit (edit some git commit messages for clarity, and also it's better for future history to have your first change and your amendments following the code review, and my final adjustments, all in 1 single commit, see https://github.com/vorburger/MariaDB4j/pull/78/commits/317347453012bcb65575f3b93473f54e85edcf02), and have that on https://github.com/vorburger/MariaDB4j/pull/78 now, thus closing this, will merge that one instead ASAP...

vorburger:
The work contributed here finally got merged to master in https://github.com/vorburger/MariaDB4j/pull/81 today.

vorburger:
The new 10.1.23 mariadb binaries contributed here just got released to Maven central (but not the MariaDB4j Java code for the export/dump; that's merged to master, and will be part of the next release 2.3.0, once I got a few more things in)

Issue #45, opened 2016-07-27
ch.vorburger.exec into separate project (with MariaDB4j depending on it)


vorburger:
The meanwhile quite mature and battle tested [ch.vorburger.exec](https://github.com/vorburger/MariaDB4j/tree/master/mariaDB4j-core/src/main/java/ch/vorburger/exec) helpers (with [the tests here](https://github.com/vorburger/MariaDB4j/tree/master/mariaDB4j-core/src/test/java/ch/vorburger/exec)) would actually make for a nice new project living separately from MariaDB4j (which would then depend on it, of course) ....

This is a draft documentation for its README: _Launching external processes from Java using the raw java.lang.ProcessBuilder API directly can be a little cumbersome.  The [Apache Commons Exec library][1] makes it a little easier.  The [ch.vorburger.exec library][2] further extends upon Commons Exec to make it truly convenient:_

```
 ManagedProcess proc = new ManagedProcessBuilder("path-to-your-executable-binary")
     .addArgument("arg1")
     .addArgument("arg2")
     .setWorkingDirectory(new File("/tmp"))
     .setDestroyOnShutdown(true)
     .setConsoleBufferMaxLines(7000)
     .build();

proc.start();
int status = proc.waitForExit();
int status = proc.waitForExitMaxMsOrDestroy(3000);
String output = proc.getConsole();

proc.startAndWaitForConsoleMessageMaxMs("started!", 7000);
// use service offered by external process...
proc.destroy();
```

Please comment / upvote if you would find this interesting & useful?

vorburger:
I could also first re-try to contribute ch.vorburger.exec into Apache Commons .. even though there never has been any reaction to https://issues.apache.org/jira/browse/EXEC-88 ...

And of course there is also https://github.com/zeroturnaround/zt-exec. But competition is healthy! ;-)

vorburger:
Started this week-end with this, see new split out repo https://github.com/vorburger/ch.vorburger.exec (which still I need to release to Maven central), and then https://github.com/vorburger/MariaDB4j/compare/ch.vorburger.exec?expand=1 ...

vorburger:
done

PR #85, opened 2017-07-15
Pull executable utility classes from new ch.vorburger.exec dependency

vorburger:
from new forked out https://github.com/vorburger/exec/ project

see https://github.com/vorburger/MariaDB4j/issues/45


vorburger:
If this fails to build, then it's because I've only JUST deployed https://github.com/vorburger/ch.vorburger.exec to Maven central - it takes a moment to get propagated .. I'll recheck, and close https://github.com/vorburger/MariaDB4j/issues/45, and then re-kick this.

PR #86, opened 2017-07-23
 Switch .travis.yml from oraclejdk7 to openjdk7

vorburger:
As mvn fails on https://travis-ci.org/vorburger/MariaDB4j/builds/256517304 :

$ javac -J-Xmx32m -version

javac 1.8.0_131

$ mvn -f DBs/pom.xml clean install

Error: JAVA_HOME is not defined correctly.

  We cannot execute /usr/lib/jvm/java-7-oracle/bin/java

The command "mvn -f DBs/pom.xml clean install" failed and exited with 1 during .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,12 @@
 language: java
-# We need Java 7, even though the source code is still Java 6, because Checkstyle during build is Java 7+ only
+
+# We need Java 7, even though the source code is still Java 6, 
+# because Checkstyle during build is Java 7+ only; and we use
+# OpenJDK instead of Oracle JDK because ya all should only
+# ever use OpenJDK!  (And get Red Hat's support for it, if
+# you use it professionally.)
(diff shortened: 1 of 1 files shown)


vorburger:
@cortiz FYI --- for some reason what you contributed in https://github.com/vorburger/MariaDB4j/commit/c3cde0c34336f868aeda3c3cfcad6f1dc89028ae does not work on Travis CI .. if you have the time to investigate why and how to fix this, a follow-up PR which re-enables it and get a PR to pass would be great!

vorburger:
Actually no, this wasn't https://github.com/vorburger/MariaDB4j/commit/c3cde0c34336f868aeda3c3cfcad6f1dc89028ae that broke Travis CI - it seems it's just generally broken; see https://travis-ci.org/vorburger/MariaDB4j/builds/256525668 ... :-(

Issue #99, opened 2018-01-11
GitHub 100 MB file limit prevents contributing latest MariaDB 10.2.11 linux x64


lde-avaleo:
Hi,

I was trying to port MariaDB 10.2.11 for win32 and linux x64. I 've followed guides and succeeded with win32. However it seems that mysqld file for linux x64 is bigger than 100MB which is github limit for single repository file.

Have you got any idea how to deal with that? Maybe we could replace DB's content to some zip files that would be unpacked and attached during maven build?

best regards,
lde

vorburger:
@lde-avaleo could you point me to where I can download a MariaDB for Linux x64 where the mysqld file is bigger than 100 MB? I've just had a look at the old ones in this repo: both 10.1.13 and 10.1.23 were about 95 MB.. I admit I actually had never even [realized that GitHub had this limit](https://help.github.com/articles/what-is-my-disk-quota/), and [that you hit this error](https://help.github.com/articles/conditions-for-large-files/)... hm.

Apparently [Git Large File Storage](https://help.github.com/articles/versioning-large-files/) could help with this... but I haven't really completely understood yet how that works. If you are willing to explore that and tell me what you want me to do, then I'm all for it! :smiley: Have you tried [installing the Git LFS client](https://git-lfs.github.com), and see if with that you can push to your repo and then raise a PR? Will I have to do anything in particular? There is nothing to "enable" in Settings that I can see. Will people who do not have the LFS client installed still be able to clone to contribute? If not, I wonder if this may be the time to finally split the MariaDB4j core Java code from the DBs into a completely separate new Git repo...

@jai-deep I think this discussion will interest you as well?

jai-deep:
@lde-avaleo  I actually tried a solution similar to your proposal some time back. I added a zip file to the repo and tried to unzip at the time of unpacking of the db. But I somehow got stuck when I got an error similar to this : https://github.com/vorburger/MariaDB4j/issues/42 . I didn't get time to work on this last week , will try this out again. You should try this out too, may be the way you implement this do the trick for us.

jai-deep:
@vorburger I explored git-lfs, basically it changes large file in your repo with the pointer to that file that will be on git-lfs server. So its works perfectly fine for a person who has git-lfs client installed. But if he do not have it , his build will fail as he only got pointer files in his repo instead of the original files. To handle this we can have a instruction file in the repo , so that a person new to the project can build this. But yes, it will add little overhead for him.

[...]

PR #100, opened 2018-01-14
Feature/maria db 10.2.11

lde-avaleo:
Port of mariaDB 10.2.11 for linux and windows.

Changed way of attaching binaries to zip file as mentioned in https://github.com/vorburger/MariaDB4j/issues/99.

Tested on Windows 10 and  Linux, version 3.10.0-514.21.1.el7.x86_64

best regards,
lde


vorburger:
@jai-deep what do you think about this?

@lde-avaleo I originally understood you want to deploy ZIP to Maven and have a run-time unzip (using some Java uncompression library), but you're doing this at build-time... that seems like a smart idea, to me! Does anyone else watching this repo have any feedback? Anyone wants to +1 / thumbs up this approach?

lde-avaleo:
I made changes regarding your comments. Build passed locally but not on travis-CI.

It failed due to permission denied OS error. Any ideas what could cause that?

vorburger:
@lde-avaleo the problem isn't TravisCI - this PR fails for me locally as well (on Linux)... But I think I see the problem... check out how in `ch.vorburger.mariadb4j.DB.unpackEmbeddedDb()` we do `Util.forceExecutable(newExecutableFile("bin", "mysql_install_db"));` but in your new MariaDB the `mysql_install_db` is now in `scripts/` instead of in `bin/` ? (BTW: The proposed https://github.com/vorburger/MariaDB4j/pull/102 would make this clearer in the log; OK for you?)

I've just pulled a mariadb-10.2.12 from https://downloads.mariadb.org and see that it indeed `mysql_install_db` is (now?) in `scripts` and no longer in `bin`, so I'm guessing you probably will want to add an `Util.forceExecutable(newExecutableFile("scripts", "mysql_install_db"));` after the `"bin", "mysql_install_db"` in `DB` as part of this? :smiley: 

BTW: Can you confirm that the ZIP you are contributing here is an unaltered original downloads.mariadb.org, without any files moved or viruses added by you? :smiling_imp:

pom.xml
```diff
@@ -38,16 +38,16 @@
 
 	<properties>
 		 <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		 <spring.version>4.3.9.RELEASE</spring.version>
-		 <springboot.version>1.5.1.RELEASE</springboot.version>
+		 <spring.version>4.3.13.RELEASE</spring.version>
```
vorburger: in an ideal world, this and following changes here should be contributed in a separate Pull Request instead of as part of this new DB... I don't want to be too much of a pain, but if you don't mind to remove this from this change (PR) and raise a new one, that would be great. PS: I'll "squash" all of your changes in this PR into a single commit upon merge, not several individual ones...

lde-avaleo: I'm sorry, I had to misunderstood what you pointed out in https://github.com/vorburger/MariaDB4j/pull/100#pullrequestreview-89912353. I have just followed https://github.com/vorburger/MariaDB4j#release.

Of course I will remove that one then.

lde-avaleo:
Well. something must changed as it is working perfectly with 2.2.3 version. I'll investigate that.

lde-avaleo:
You're right. I was missing chenge to executable scripts/mysql_install_db. It seems like my linux CI agents have different default user permissions thats why it works for us. It sems like build is green now.

I confirm that binaries come from original content of MariaDB linux dist.

vorburger:
@lde-avaleo thank you! LGTM, and I'm happy to merge this now (and nobody replied to my request for feedback above 1 week ago), but if you have the motivation, I would have 1 last suggestion... Motivated by the the GitHub git push warning for the 50 MB ZIP file size (which you probably also noticed), I just did a quick local test using other compression formats, and found that both XZ and 7z would give us 30 MB archives instead of the 56.8 MB ZIP for the Linux one. 

It's not super super important, so if you've had enough just say so and I'll merge, but as we get more of these updates for new mariadb releases, then saving ca. 25 MB in the git repo each time (these stay "forever", of course) by having a XZ or 7z instead of the ZIP in the repo, and changing the Maven antrun accordingly, could be... nice.

FYI Ant (via Maven antrun) doesn't seem to have a native unxz or un7z like it has an unzip, but [according to this StackOverflow post it does not seem to be very hard](https://stackoverflow.com/questions/31603742/untar-unzip-tar-xz-with-ant)? But [note here that the latest ant-compress seems to be 1.5 now and no longer 1.4](http://ant.apache.org/antlibs/compress/).

lde-avaleo:
Can we create separate feature task for that? Unfortunately my timebox for that has just ended up. :(

vorburger:
@lde-avaleo sure, no problem; merged! Thanks again for your contribution. If you have the time to follow-up with more contributions, e.g. for the version bump, that's cool. Don't worry about changing the archive format for this one anymore, once it's merged in Git and Maven deployed it's too late; but perhaps for future new mariaDB4 versions I'll suggest to use XZ or 7z instead of ZIP.

FYI I've just deployed mariaDB4j-db-linux64-10.2.11 & mariaDB4j-db-win32-10.2.11 to Maven central, to avoid https://github.com/vorburger/MariaDB4j/issues/101. (It can take a few hours to propagate to their CDN.)

PR #102, opened 2018-01-24
chmod +x now prints info level log if file does not exist

vorburger:
motivated by analysis of the problem in https://github.com/vorburger/MariaDB4j/pull/100

@lde-avaleo if this seems like a good idea to you as well, can you reply with +1 to this PR?


vorburger:
Overriding as the missing approval is due to VersionEye.com's shutdown, I've just removed it; snif!

Issue #112, opened 2018-03-25
Enable multiple srcripts to be sourced like flyway:migrate


usmanakram232:
Enable flyway style migration scripts to be sourced.

I propose, we add a new method which take the path to directory containing migration scripts and sources them in order.

``` java
  public void migrate(String directory, String suffix)
```

usmanakram232:
As, discussed in PR #113 instead of trying to replicate flyway:migrate. We can have a `sourceAll` method with loads all the script in given directory.

vorburger:
@usmanakram232 sounds good to me (and @duttonw seems to like this as well).

More FTR in #113 this is going to be named `sourceAll()` not `migrate()` - that makes sense.

Once #113 is merged, can this issue be closed, or will there be anything else to do in this issue?

usmanakram232:
@vorburger yes, we can close this issue upon merge,

vorburger:
@usmanakram232 hope you don't mind if I close this issue, we have the active #113 PR to track it now.

Issue #106, opened 2018-02-26
Release 2.3.0


vorburger:
https://github.com/vorburger/MariaDB4j/issues/105 is a reminder that I **really** should deploy a new releast to Maven central ASAP...

vorburger:
I'll try to finally do a new release this week, but before that would like to do the following:

* @neeleshs or anyone else wrap up https://github.com/vorburger/MariaDB4j/pull/96
* fix the  https://github.com/vorburger/MariaDB4j/issues/88 bug re wrong use of annotations
* new dump API should not log entire DB [exec setStdOutLogging(false)]
* update the [release notes in CHANGES](https://github.com/vorburger/MariaDB4j/blob/master/CHANGES.md)

Would anyone using MariaDB4j and watching this repo reading this be willing to help with any of the above?

LoZeno:
We very recently started using MariaDB4J and we could really use the setSecurityDisabled() feature, but I have no idea how to help to deploy a new release of anything to maven central.
So I wonder, would you consider "paying you with pizza" a suitable help? :)

cyzhao:
@vorburger any update on this issue? Please let me know what I can help with.

vorburger:
I've just pushed 2.3.0 to Maven central, and it should show up there shortly (takes a few hours):

```
    (...)
    Uploading to ossrh: https://oss.sonatype.org:443/service/local/staging/deployByRepositoryId/chvorburger-1039/ch/vorburger/mariaDB4j/mariaDB4j-app/2.3.0/mariaDB4j-app-2.3.0.pom.asc
    Uploaded to ossrh: https://oss.sonatype.org:443/service/local/staging/deployByRepositoryId/chvorburger-1039/ch/vorburger/mariaDB4j/mariaDB4j-app/2.3.0/mariaDB4j-app-2.3.0.pom.asc (455 B at 2.4 kB/s)
    Uploading to ossrh: https://oss.sonatype.org:443/service/local/staging/deployByRepositoryId/chvorburger-1039/ch/vorburger/mariaDB4j/mariaDB4j/2.3.0/mariaDB4j-2.3.0-sources.jar
    Uploaded to ossrh: https://oss.sonatype.org:443/service/local/staging/deployByRepositoryId/chvorburger-1039/ch/vorburger/mariaDB4j/mariaDB4j/2.3.0/mariaDB4j-2.3.0-sources.jar (3.0 kB at 4.9 kB/s)
[...]

PR #96, opened 2017-12-26
Issue #63 - Provide a callback if the DB process crashes

neeleshs:
Sorry for the delay on this. Got distracted with work

PR #113, opened 2018-03-25
Enables  sourcing multiple sql script form the given directory 

usmanakram232:
This *PR* includes changes for #112 .

## This PR includes following changes:
Enable DB to source all of sql migration scripts in simple order.

## Changes are 
 - [ ] Breaking Changes (API/Behavior change)
 - [X] New Feature
 - [ ] Bug fix


mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/Util.java
```diff
@@ -175,4 +177,36 @@ private static void tryN(int n, long msToWait, Procedure<IOException> procedure)
...
+     * @return
+     */
+    public static List<String> getFileList(String directory, String suffix) {
+        List<String> files = new ArrayList<String>();
+        File dir = new File(directory);
+        if (dir.isDirectory()){
+            File[] filesList = dir.listFiles();
+            if (filesList == null)
+                throw new IllegalArgumentException(
+                    "Unable to retrive list of ilies in: " + directory);
```
vorburger: what does "ilies" mean in this error message? Typo?

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/Util.java
```diff
@@ -175,4 +177,36 @@ private static void tryN(int n, long msToWait, Procedure<IOException> procedure)
...
+     * Retrieves sorted list of files from $directory/*.suffix
+     *
+     * @param directory directory within class path from where files are to be retrieved
+     * @param suffix to filter file type
+     * @return
+     */
+    public static List<String> getFileList(String directory, String suffix) {
+        List<String> files = new ArrayList<String>();
+        File dir = new File(directory);
+        if (dir.isDirectory()){
```
vorburger: nit pick: missing space before {

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/Util.java
```diff
@@ -175,4 +177,36 @@ private static void tryN(int n, long msToWait, Procedure<IOException> procedure)
...
+            File[] filesList = dir.listFiles();
+            if (filesList == null)
+                throw new IllegalArgumentException(
+                    "Unable to retrive list of ilies in: " + directory);
+
+
+            for (File f: filesList) {
+                if (f.isFile() && f.getName().endsWith(suffix)){
```
vorburger: nit pick: missing space before {

vorburger:
The question is whether this should be part of the core library. Personally I have no big objections to this, because it does not actually introduce a dependency to FlyWay, it just adds some "convenience" functions, based on a convention. Do other contributors and MariaDB4j users have any thoughts if this is OK with them?

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/Util.java
usmanakram232: Fixed, Thanks.

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/Util.java
```diff
@@ -175,4 +177,36 @@ private static void tryN(int n, long msToWait, Procedure<IOException> procedure)
...
+            for (File f: filesList) {
+                if (f.isFile() && f.getName().endsWith(suffix)) {
+                    files.add(f.getName());
+                }
+            }
+
+            java.util.Collections.sort(files);
+        } else {
+            throw new IllegalArgumentException(
+                "Path does not points to a directory: " + directory);
```
vorburger: "does not point" not "points"

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
```diff
@@ -224,14 +224,48 @@ protected File getAbsoluteSocketFile() {
...
+   */
+    public void migrate(final String directory, final String suffix)
+        throws ManagedProcessException {
+        String directoryPath = getClass().getClassLoader()
+            .getResource(directory).getFile();
+        List<String> files = Util.getFileList(directoryPath, suffix);
+
+        if (files == null)
+            throw new IllegalArgumentException(
+                "Unable to retrive list of ilies in: " + directory);
```
vorburger: still "ilies" instead of "files" here (you fixed it only elsewhere, in Utils)

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
```diff
@@ -224,14 +224,48 @@ protected File getAbsoluteSocketFile() {
...
+   * and sources them via the mysql command line tool.
+   *
+   * @param directory path that contain sql migration scripts to load into db
+   * @param suffix suffix to filter script to load
+   * @throws ManagedProcessException if something fatal went wrong
+   */
+    public void migrate(final String directory, final String suffix)
+        throws ManagedProcessException {
+        String directoryPath = getClass().getClassLoader()
+            .getResource(directory).getFile();
```
vorburger: I'm wondering if there isn't a problem with this approach: I think (?) this actually only works for physical directories, not for files contained inside a JAR on the classpath (that is not "unpacked" to a directory).  

This is different from how `DB.source()` itself works, which this calls, which seems to only support sourcing from the classpath, and that inconsistency I think will create confusion for users.

Thinking further about this, I'm realizing that perhaps what you really want here is not so much a `migrate()` but really more just something quite like the existing `source()` method(s!), where instead of a `String resource` _path to a resource on the classpath to source_, one could just pass a "resourceBase" and it would _find all resources on the classpath in that base (which may be a directory, but does not always have to be...)_ and source them? Have you seen how the `ch.vorburger.mariadb4j.Util.extractFromClasspathToFile(String, File)` uses the `ResourcePatternResolver` (from the Spring Framework) to find all resources on a location - how about also using that here? So maybe just:

    sourceAll(String packagePath)

as well as the same with `String username, String password, String dbName` (like `source()` has).

You could still also have a `String suffix` (for .sql), if that's really needed.. but is it, really; wouldn't one typically have such sql scripts for migrations, or other, alone?

Could you also add a test for the new method?

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/Util.java
```diff
@@ -175,4 +177,36 @@ private static void tryN(int n, long msToWait, Procedure<IOException> procedure)
...
+
+    /**
+     * Retrieves sorted list of files from $directory/*.suffix.
+     *
+     * @param directory directory within class path from where files are to be
+     * retrieved
+     * @param suffix to filter file type
+     * @return List of files in the directory
+     */
+    public static List<String> getFileList(final String directory, final String suffix) {
```
vorburger: if we go for the idea above, then perhaps we can just completely rm this entire new method, as it won't be used then?

usmanakram232: I will remove it.

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
vorburger: The ordering problem is interesting... I wonder how everyone else out there using Flyway does this correctly?

vorburger:
@usmanakram232 I'm about to release 2.3.0 (#106) and unless you, or anyone else reading this, want to do what we discuss above ASAP I don't think this will make it into that release...

vorburger:
@usmanakram232 it looks like you lost interest in this? Are there any other others in the MariaDB4j user community who are willing to step to finish this (as in #96) ? Or shall we just close this PR at this point - thoughts, anyone, re. the value of this idea? @duttonw @yiftizur @thesquaregroot

usmanakram232:
~~I totally forgot about it, as I had be traveling/busy~~ If this feature seems useful, I can make changes as you suggested; otherwise we can close it. As, for my use case a simple implementation might not work and I don't think it is the place to do much processing on file names. 

So, personally I am using flyway to migrate test db.

duttonw:
So is this something to emulate part of  https://flywaydb.org/documentation/command/migrate ?
The convention of flyway running an ordered list of vanilla sql files could be done but it won't handle if the database is stored outside of temp directory (flyway does a versioning table inside the database and compares if it had already run said scripts, simliar to dbdeploy and https://www.liquibase.org/. 
Having an option to load multiple large sql files sounds good. But we should not mention anything about flyway as it won't handle re-runs etc.

Now from my research, flyway has a maven plugin, and if ordered after the recent mariaDB4j plugin (maven is order bound in its xml) it will have the database started and its variables for where that database lives provided to it and then you have the full support of flyway (and if you pay for pro/enterprise you get more features).
https://flywaydb.org/documentation/maven/ https://github.com/flyway/flyway/tree/master/flyway-maven-plugin

So this work could be included we just need to get more tests to ensure it works correctly and enhance  the maven mariaDB4j plugin to allow a folder scan option and use that function call that was created.

vorburger:
@usmanakram232 thanks for picking this up! The various unrelated code formatting changes that this PR has together with the new functionality make this a little harder for me to quickly but safely review than it could (should) be... I don't suppose you could remove all of those, so that this PR only shows what actually changed? Either just revert them, or re-raise them in a "formatting only changes" separate PR. Just a thought. If not possible, I'll get back to looking over this when I have a bit more time later this week.

@duttonw given your interest, I would welcome any review comments you have on this PR as well!

usmanakram232:
One important thing before we merge. As, we discussed earlier, now I use 
``` java
ch.vorburger.mariadb4j.Util.extractFromClasspathToFile(String, File)
```
to extract the scripts but where should be the target dir? Currently using, `$baseDir/sources/$inputPath`. I am not sure, if it is appropriate target for extraction.

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleTutorialTest.java
```diff
@@ -84,6 +84,49 @@ public void testEmbeddedMariaDB4j() throws Exception {
...
+            qr.update(conn, "INSERT INTO hello VALUES ('Hello, world')");
+
+            // Should be able to select from a table
+            List<String> results = qr.query(conn, "SELECT * FROM hello",
+                    new ColumnListHandler<>());
+            Assert.assertEquals(1, results.size());
+            Assert.assertEquals("Hello, world", results.get(0));
+
+            // Should be able to source a SQL file
+            db.sourceAll(".", "sql", "root", null, dbName);
```
vorburger: you could `assertThat(db.sourceAll()).isEqualTo(1) here? Assuming there is 1 *.sql file - or 3 or whatever, I didn't check.

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/Util.java
```diff
@@ -175,4 +180,24 @@ private static void tryN(int n, long msToWait, Procedure<IOException> procedure)
...
+     */
+    public static List<Path> getFileList(final Path directory, final String suffix)
+        throws IOException {
+        List<Path> files = new ArrayList<>();
+        try (DirectoryStream<Path> pathStream =
+            Files.newDirectoryStream(directory, path -> path.toFile().isFile() && path.endsWith(suffix))) {
+          pathStream.forEach(entry -> files.add(entry));
+        }
+        return files;
+    }
```
vorburger: this method does not seem to do any sorting - contrary to what it's JavaDoc claims ;-) as such doesn't this make it impossible to use it e.g. with Flyway, or anything else like that? I'm not saying that this should have hard-coded fixed sorting, but in earlier review discussions you menioned want to pass a Comparator... that would seem seemful? Either has a parameter her (and adapt JavaDoc), or otherwise make it clear that this is unsorted (return a Set instead of a List?) and let sourceAll() accept a Comparator and apply it to what this returns?

vorburger:
@usmanakram232 @duttonw I came across this as I'm preparing to #165 ... I'll release without this, but should (either of) you still have any interest this feature, please don't let this discourage you - I'm still very happy to merge this after the release if you would like to address the last review comment above.

PR #126, opened 2018-05-15
compile with Java 8 instead of 6

vorburger:
and run travis with 8 instead of 7

because Spring Boot 2 seems to require Java 8 (?)


vorburger:
does anyone watching this repo have any objections to going from Java 6 to Java 8 here?

PR #131, opened 2018-05-15
remove <version>1.5.5 from mariadb-java-client which bumps it to 2.2.3

vorburger:
The artifact is managed in
org.springframework.boot:spring-boot-dependencies:2.0.2.RELEASE


vorburger:
This was a "counter proposal" to #124.

PR #133, opened 2018-06-13
Only set --max_allowed_packet if it is not in the configured arguments

thesquaregroot:
In one of our projects, we were planning to change the max_allowed_packet parameter but since this was automatically being set we wanted to confirm that our value would override the 64M value set automatically by MariaDB4j.

I had difficulty finding documentation confirming that the expected behavior of repeated command-line arguments is that the last value will be used.  In addition, having this set multiple times to different values could generally cause confusion.

This change simply checks the arguments that have been configured to see if it has already been set and only adds the parameter if it will not appear later.


vorburger:
Thanks for your (first) contributions - more welcome any time.. ;-)

PR #139, opened 2018-07-13
Add MariaDB4J as JUnit Rule

yiftizur:
Add MariaDB4j as an External Resource to be used as a JUnit rule, with some convenience methods for initializing the DB.

### DIFF
--- a/mariaDB4j-core/pom.xml
+++ b/mariaDB4j-core/pom.xml
@@ -45,7 +45,6 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.springframework</groupId>
--- a/mariaDB4j-core/pom.xml
+++ b/mariaDB4j-core/pom.xml
@@ -45,6 +45,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>org.springframework</groupId>
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfiguration.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfiguration.java
@@ -66,6 +66,8 @@ public interface DBConfiguration {
     /** Whether to to "--skip-grant-tables". */
     boolean isSecurityDisabled();
 
+    String getConnectionURL(String dbName);
+
     static class Impl implements DBConfiguration {
 
         private final int port;
@@ -152,6 +154,10 @@ public interface DBConfiguration {
             return isSecurityDisabled;
         }
 
+        @Override
+        public String getConnectionURL(String dbName) {
+            return "jdbc:mysql://localhost:" + getPort() + "/" + dbName;
+        }
     }
 
 }
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfiguration.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfiguration.java
@@ -20,6 +20,7 @@
 package ch.vorburger.mariadb4j;
 
 import java.util.List;
+import java.util.function.Function;
 
 /**
  * Enables passing in custom options when starting up the database server This is the analog to
@@ -66,7 +67,7 @@ public interface DBConfiguration {
     /** Whether to to "--skip-grant-tables". */
     boolean isSecurityDisabled();
 
-    String getConnectionURL(String dbName);
+    String getURL(String dbName);
 
     static class Impl implements DBConfiguration {
 
@@ -81,10 +82,11 @@ public interface DBConfiguration {
         private final List<String> args;
         private final String osLibraryEnvironmentVarName;
         private final boolean isSecurityDisabled;
+        private final Function<String, String> getURL;
 
         Impl(int port, String socket, String binariesClassPathLocation, String baseDir, String libDir, String dataDir,
-                boolean isWindows, List<String> args, String osLibraryEnvironmentVarName, boolean isSecurityDisabled,
-                boolean isDeletingTemporaryBaseAndDataDirsOnShutdown) {
+             boolean isWindows, List<String> args, String osLibraryEnvironmentVarName, boolean isSecurityDisabled,
+             boolean isDeletingTemporaryBaseAndDataDirsOnShutdown, Function<String, String> getURL) {
             super();
             this.port = port;
             this.socket = socket;
@@ -97,6 +99,7 @@ public interface DBConfiguration {
             this.args = args;
             this.osLibraryEnvironmentVarName = osLibraryEnvironmentVarName;
             this.isSecurityDisabled = isSecurityDisabled;
+            this.getURL = getURL;
         }
 
         @Override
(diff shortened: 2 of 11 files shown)


mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfiguration.java
```diff
@@ -152,6 +154,10 @@ public boolean isSecurityDisabled() {
             return isSecurityDisabled;
         }
 
+        @Override
+        public String getConnectionURL(String dbName) {
+            return "jdbc:mysql://localhost:" + getPort() + "/" + dbName;
```
vorburger: did you notice that this exact same method is already available (as `getURL()`) on the `DBConfigurationBuilder` ? The existing tests use that. Perhaps we can avoid confusion by not adding this 2nd identical method also to the `DBConfiguration`, and you can change your test to use that? 

Or if you do this in the `MariaDBRule` where you have only a DBConfiguration and not a DBConfigurationBuilder (which seems fine), then keep it, but factor out and re-use instead of copy/paste the `"jdbc:mysql://localhost:" + getPort() + "/" + dbName;` ? Not that it will ever change, but still... :smile: Also let's name just this `getURL()` without *Connection*, just for consistency with the existing one?

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
```diff
@@ -0,0 +1,76 @@
...
+import ch.vorburger.exec.ManagedProcessException;
+import ch.vorburger.mariadb4j.DB;
+import ch.vorburger.mariadb4j.DBConfiguration;
+import ch.vorburger.mariadb4j.DBConfigurationBuilder;
+import org.junit.rules.ExternalResource;
+
+public class MariaDBRule extends ExternalResource {
+
+    private DB db;
+    private DBConfiguration config;
```
vorburger: also `final`

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    protected void before() throws Throwable {
+        db = DB.newEmbeddedDB(config);
+        db.start();
+        initDB();
+    }
+
+    private void initDB() throws ManagedProcessException {
```
vorburger: make this `protected` instead of `private` in case users want to extend?

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
```diff
@@ -0,0 +1,76 @@
...
+        if(resource != null)
+            db.source(resource, dbName);
+    }
+
+    @Override
+    protected void after() {
+        try {
+            db.stop();
+        } catch (ManagedProcessException e) {
+            throw new AssertionError();
```
vorburger: loosing the `e` here will make it hard to understand the root cause, so  use `throw new AssertionError("db.stop() failed", e)` instead.

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    public String getConnectionString() {
```
vorburger: how about just naming this `getURL()` instead? (Or `getConnectionURL()` if getURL is too short; but it would be consistent with the existing `getURL` on DBConfigurationBuilder.) PS: I know it's a String and not URL type, but get*String() seems ugly, to me... :-)

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
```diff
@@ -0,0 +1,76 @@
...
+        return db.getConfiguration().getConnectionURL(dbName);
+    }
+
+    public DBConfiguration getConfiguration() {
```
vorburger: wanna name this getDBConfiguration while we at it? Just a thought.

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+package ch.vorburger.mariadb4j.junit;
+
+import ch.vorburger.exec.ManagedProcessException;
+import ch.vorburger.mariadb4j.DB;
+import ch.vorburger.mariadb4j.DBConfiguration;
+import ch.vorburger.mariadb4j.DBConfigurationBuilder;
+import org.junit.rules.ExternalResource;
+
+public class MariaDBRule extends ExternalResource {
```
vorburger: wanna name this the MariaDB**4j**Rule? Just a thought.

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
yiftizur: 👍

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
yiftizur: Excellent idea!

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
yiftizur: 👍

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
yiftizur: will do

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
yiftizur: 👍

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/junit/MariaDBRule.java
yiftizur: 👍

mariaDB4j-core/pom.xml
```diff
@@ -45,7 +45,6 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <scope>test</scope>
```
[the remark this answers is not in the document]
yiftizur: I was not familiar with the optional dependency option, i really like it :) will change it to that

yiftizur:
Hey @vorburger, thanks for all the feedback!
I've made the changes you requested, so whenever you want you can review them :)

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfiguration.java
```diff
@@ -79,10 +82,11 @@
         private final List<String> args;
         private final String osLibraryEnvironmentVarName;
         private final boolean isSecurityDisabled;
+        private final Function<String, String> getURL;
```
vorburger: nice! :smile:

PR #153, opened 2018-08-17
Add mariaDB4j-springboot module for auto-configuration with spring boot

kbyyd24:
Add a module named mariaDB4j-springboot, it will auto-configure `MariaDBSpringService` as a bean named `mariaDB4j`, and initialize bean `dataSource` after `mariaDB4j` initialized by annotation `@DependsOn`.
So now, we can use mariaDB4j with spring boot with dependency of mariaDB4j-springboot. For example, we can configure gradle like this:

```gradle
dependencies {
    compile 'org.springframework.boot:spring-boot-starter'
    testCompile 'ch.vorburger.mariaDB4j:mariaDB4j-springboot:2.3.1-SNAPSHOT'
}
```

### DIFF
--- a/mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
+++ b/mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
@@ -1,0 +1,46 @@
+/*
+ * #%L
+ * MariaDB4j
+ * %%
+ * Copyright (C) 2012 - 2014 Michael Vorburger
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package ch.vorburger.mariadb4j.springboot.autoconfigure;
+
+import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.jdbc.DataSourceBuilder;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.DependsOn;
+
+import javax.sql.DataSource;
+
+//separate with MariaDB4jSpringConfiguration for test of it
+@Configuration
+@ConfigurationProperties("spring.datasource")
+public class DataSourceAutoConfiguration {
+
+    @Bean
+    @DependsOn("mariaDB4j")
+    public DataSource dataSource(DataSourceProperties dataSourceProperties) {
+        return DataSourceBuilder.create()
+                .driverClassName(dataSourceProperties.getDriverClassName())
+                .url(dataSourceProperties.getUrl())
+                .username(dataSourceProperties.getUsername())
+                .password(dataSourceProperties.getPassword())
+                .build();
+    }
+}
--- a/mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/MariaDB4jSpringConfiguration.java
+++ b/mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/MariaDB4jSpringConfiguration.java
@@ -1,0 +1,36 @@
+/*
+ * #%L
+ * MariaDB4j
+ * %%
+ * Copyright (C) 2012 - 2014 Michael Vorburger
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package ch.vorburger.mariadb4j.springboot.autoconfigure;
+
+import ch.vorburger.mariadb4j.springframework.MariaDB4jSpringService;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConfigurationProperties
+public class MariaDB4jSpringConfiguration {
+
+    @Bean
+    public MariaDB4jSpringService mariaDB4j() {
+        return new MariaDB4jSpringService();
+    }
+
+}
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -21,6 +21,7 @@ MariaDB4j CONTRIBUTORS
 - Andrew Groot [@thesquaregroot](https://github.com/thesquaregroot) <groot@softwareverde.com>, June 2018; for https://softwareverde.com
 - Yftach Zur [@yiftizur](https://github.com/yiftizur), June 2018; for MariaDB4jRule for easy integration with JUnit
 - William Dutton [@duttonw](https://github.com/duttonw)<will.dutt@gmail.com>, June 2018; for mariaDB4j-maven-plugin integration testing of micro services under services.qld.gov.au
+- Yuexiang Gao [@kbyyd24](https://github.com/kbyyd24) <melo@gaoyuexiang.cn> (http://blog.gaoyuexiang.cn), August 2018; for mariaDB4j-springboot auto-configure with spring boot 
 - also see https://github.com/vorburger/MariaDB4j/graphs/contributors
 
 Contributions, patches, forks more than welcome - hack it, and add your name here! ;-)
--- a/mariaDB4j-springboot/pom.xml
+++ b/mariaDB4j-springboot/pom.xml
@@ -1,0 +1,37 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>ch.vorburger.mariaDB4j</groupId>
+        <artifactId>mariaDB4j-pom</artifactId>
+        <version>2.3.1-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>mariaDB4j-springboot</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>mariaDB4j</artifactId>
+            <version>${project.version}</version>
+        </dependency>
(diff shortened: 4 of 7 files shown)


duttonw:
can you update the readme in the root on how to use this as well as create an example integration project to has sample config setup with spring boot. (how you will use this in a simple spring-boot app)
see https://github.com/spring-projects/spring-boot/tree/master/spring-boot-samples/spring-boot-sample-flyway but for MariaDB4j instead.

kbyyd24:
@duttonw I will update the readme in the root. But it's necessary to create another example integration project to has sample config setup with spring boot? I think mariaDB4j-app is a sample of it, maybe what I need to do is update it. Is it OK to change mariaDB4j-app?

vorburger:
>  I think mariaDB4j-app is a sample of it, maybe what I need to do is update it. Is it OK to change mariaDB4j-app?

That makes sense to me, yes - go for it.

vorburger:
@kbyyd24 if you work @duttonw feedback into this and resolve the minor new merge conflict, I'd review this.

kbyyd24:
@vorburger I updated readme, almost like @duttonw feedback, the only thing different is the link to the  example application, I changed it to [mariaDB4j-app](https://github.com/vorburger/MariaDB4j/blob/master/mariaDB4j-app/). I Also resolved the conflict. So I think it's ready to be review.

kbyyd24:
@duttonw Thanks for your feedback about the readme and my blog. I have updated the readme as you feedback. For the JHipster, I had never heard about it, I will try it for a demo. Thanks for your recommend.
By the way, how can you read my blog? It Chinese, do you know that?

duttonw:
Hi @kbyyd24 ,

Google chrome allows language translation with a click of a button or [like this](https://translate.google.com/translate?sl=auto&tl=en&js=y&prev=_t&hl=en&ie=UTF-8&u=http%3A%2F%2Fblog.gaoyuexiang.cn%2F2017%2F10%2F25%2FCustome_Validation_in_Java%2F&edit-text=)

kbyyd24:
@duttonw Oh, I forgot it :joy:

kbyyd24:
Hi @duttonw , I don't know how to do that. I clicked the "Update branch" button, then triggered travis automatically. After pipeline passed, this page looked like the picture. Is the rebase done?

image

duttonw:
Cool, now it just needs @vorburger to do a final review and merge :D :+1:

kbyyd24:
That's cool, Thank you for your help 👍

mariaDB4j-app/pom.xml
```diff
@@ -43,6 +43,22 @@
...
+		</dependency>
+		<dependency>
+			<!-- Use JPA for example -->
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-data-jpa</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>javax.xml.bind</groupId>
+			<artifactId>jaxb-api</artifactId>
+			<version>2.3.0</version>
```
vorburger: why is this needed here?

mariaDB4j-app/pom.xml
```diff
@@ -43,6 +43,22 @@
 			<artifactId>spring-boot-starter</artifactId>
 			<optional>true</optional> <!-- !!! -->
 		</dependency>
+
+		<dependency>
+			<groupId>org.mariadb.jdbc</groupId>
+			<artifactId>mariadb-java-client</artifactId>
+			<version>2.2.6</version>
```
vorburger: I doubt this is needed, have you tried without it?

Note it currently works without this.... 

The parent/root POM already has this dependency.

Why fix a more recent version? If it were absolutely needed (I doubt it, easier to inherit it from the Spring BOM instead of having to keep bumping it here @duttonw), why only here, not in the parent?

mariaDB4j-app/src/main/java/ch/vorburger/mariadb4j/springframework/boot/MariaDB4jApplication.java
```diff
@@ -63,6 +62,7 @@ public static void main(String[] args) throws Exception {
 
     @Override
     public int getExitCode() {
+        MariaDB4jSpringService dbService = applicationContext.getBean(MariaDB4jSpringService.class);
```
vorburger: it's been many years since I last used the Spring Framework myself (doing other things nowadays), but wasn't the idea that explicitly looking up beans is bad? I thought it was much better DI practice to just inject what you needed. So why change the ` @Autowired MariaDB4jSpringService` to an `applicationContext.getBean(MariaDB4jSpringService.class)` ?

mariaDB4j-app/src/main/resources/application.yml
```diff
@@ -0,0 +1,9 @@
+spring:
+  datasource:
+    driver-class-name: org.mariadb.jdbc.Driver
+    url: jdbc:mariadb://localhost:${mariaDB4j.port}/test
+    username: root
+    password:
+
+mariaDB4j:
+  port: 3307
```
vorburger: hard-coding to port 3307 means this fails if one runs this on a system where there already is a mysqld running (e.g. from a RPM / DEB launched by systemd).  The original `mariaDB4j.port=0` thing let it automatically choose a free port - that was cool. Is it possible to retain a similar solution? Perhaps it's even enough to simply try out `port: 0` here?

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * #%L
+ * MariaDB4j
+ * %%
+ * Copyright (C) 2012 - 2014 Michael Vorburger
```
vorburger: nit picking, sorry but could you put "2012 - 2018" here?

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/MariaDB4jSpringConfiguration.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * #%L
+ * MariaDB4j
+ * %%
+ * Copyright (C) 2012 - 2014 Michael Vorburger
```
vorburger: 2018

mariaDB4j-app/pom.xml
kbyyd24: I can't find the dependency in the parent/root POM. I can only find it in `mariaDB4j` module and it's scope is test. So need I add the dependency in the parent/root POM?

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
kbyyd24: Thank you for your mention, I had never notice this 😂.
Now it fixed.

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/MariaDB4jSpringConfiguration.java
kbyyd24: Fixed.

mariaDB4j-app/src/main/resources/application.yml
duttonw: Hi @kbyyd24 , have you looked at how in-memory db's get their dynamic port?

kbyyd24: @duttonw No, I don't know how spring get that. I will research about that.

kbyyd24: Hi @duttonw , I found that h2, hsql and derby are auto configured in Spring Boot, and they are in-memory DB. So they are running in the same process with Spring Boot. Then we don't need any port to connect to them.
In this case, mariaDB4j is not in-memory DB, so we need to determine a port. I'm going to change the logic of creation dataSource bean. Maybe we don't need use property of `spring.data` here, we can use `mariaDB4j.port`, `mariaDB4j.db-host`, `mariaDB4j.username` and etc.

mariaDB4j-app/pom.xml
vorburger: see #164

mariaDB4j-app/src/main/resources/application.yml
jordanms: You should use a random port and then set the chosen port in the [datasource](https://github.com/vorburger/MariaDB4j/pull/153/files#diff-5bc8877ceebd71b5302b7e32eb8c130613b238b1f6a75a0aa84161ca267b3a7cR41).

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+@Configuration
+@ConfigurationProperties("spring.datasource")
+public class DataSourceAutoConfiguration {
+
+    @Bean
+    @DependsOn("mariaDB4j")
+    public DataSource dataSource(DataSourceProperties dataSourceProperties) {
+        return DataSourceBuilder.create()
+                .driverClassName(dataSourceProperties.getDriverClassName())
+                .url(dataSourceProperties.getUrl())
```
jordanms, line 41: You should not use the URL coming from the configuration because you will use the production config if it is not overwritten in the test config. You should set it to `jdbc:mariadb://localhost:[dynamic_chosen_port]/test`.

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+                .username(dataSourceProperties.getUsername())
```
jordanms, line 42: Same here. You should usee `root`

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+                .password(dataSourceProperties.getPassword())
```
jordanms, line 43: Same here. You should use `` (empty).

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
jordanms: Of course, has also to consider the `mariaDB4j` configuration...

vorburger:
@jordanms You would need to open a new PR with suggested changes, or a new issue describing a problem, or suggesting an enhancement. Comments on a 4 year old closed PR are ignored.

duttonw:
Hi @jordanms , 

To your comment that you were providing feedback. You need to be in the moment and that moment has past 4 years ago. 

Yes this is an opinionated way of doing it as spring boot is 'very opinionated', sometimes that opinion changes direction and falls out of flavour, but it still is a way of working. 

If you don't like how it was done in the past. Then contribute, change it in a new Pull Request or Raise an Issue ticket as an enhancement request on how to bring it into line with current day spring boot.  

I won't be contributing at this time as you have not articulated in a new ticket on how it could change in a manner that would compel another to implement it.

Regards,

@duttonw 

p.s. This could now be done with conditionals properties which would only enable this class if ```mariaDB4j.port``` property is not found. https://www.baeldung.com/spring-conditional-annotations#predefined-conditional-annotations

jordanms:
Hi, @duttonw 

> If you don't like how it was done in the past.

It's not that I don't like it, it just doesn't work. I spent hours of my day trying to figure out how to make it work. The worse thing is that before I try to use the library, I checked if it was maintained and the last update was a few days ago. So I decided to try it. 

When I say that it doesn't work, is because the idea of the feature is that you have autoconfiguration in the way of H2 for testing, and this is not done and has never been achieved from what I saw in the discussions.

I only commented because I thought it could be helpful. I'm not using the library because it uses a very old MariaDB (10.2 I think) which is not supported by Flyway.

Regards

PR #155, opened 2018-08-20
use fixed versions of maven-jar-plugin and maven-invoker-plugin

vorburger:
This fixes the Maven warnings which appear on the build:

```
[WARNING] Some problems were encountered while building the effective model for ch.vorburger.mariaDB4j:mariaDB4j-maven-plugin:maven-plugin:2.3.1-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-jar-plugin is missing. @ ch.vorburger.mariaDB4j:mariaDB4j-maven-plugin:[unknown-version], /home/vorburger/dev/MariaDB4j/git/MariaDB4j/mariaDB4j-maven-plugin/pom.xml, line 181, column 21
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-invoker-plugin is missing. @ ch.vorburger.mariaDB4j:mariaDB4j-maven-plugin:[unknown-version], /home/vorburger/dev/MariaDB4j/git/MariaDB4j/mariaDB4j-maven-plugin/pom.xml, line 259, column 29
[WARNING]
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING]
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
```

### DIFF
--- a/mariaDB4j-maven-plugin/pom.xml
+++ b/mariaDB4j-maven-plugin/pom.xml
@@ -28,7 +28,7 @@
             <artifactId>mariaDB4j</artifactId>
             <version>${project.version}</version>
         </dependency>
-       <dependency>
+        <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
         </dependency>
@@ -181,6 +181,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
+                <version>3.1.0</version>
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


vorburger:
@duttonw OK for you?

duttonw:
:+1: 
If we import spring-boot's bill of materials, it should fix these problems for us, but then we would be tied to spring's milestones/versions, e.g like this [example](https://www.baeldung.com/spring-maven-bom).

We may want to have our own BOM which is a fork/copy of springs as a base and allow dependency bot to keep up to date so we can move away from having versions in each pom.xml.

vorburger:
> If we import spring-boot's bill of materials

we are, see https://github.com/vorburger/MariaDB4j/blob/master/pom.xml#L124 ... and mariaDB4j-maven-plugin has that as its `parent`, but for some reason that did not cover these two plugins.

> We may want to have our own BOM which is a fork/copy of springs as a base and allow dependency bot to keep up to date so we can move away from having versions in each pom.xml.

we could simply put more into the `dependencyManagement` of the root parent POM as well.

vorburger:
also caf6426cfb5a68078412b2a44f14856e0c8eb8ea

PR #164, opened 2018-08-27
clean up mariaDB4j-app POM

vorburger:


vorburger:
@kbyyd24 (and @duttonw) OK for you?

vorburger:
BTW it (mariadb-java-client) just got bumped in #158, and with this change we have the version only in 1 place.

duttonw:
If it compiles, yep 👍

kbyyd24:
Yeah, that's OK

PR #190, opened 2018-11-16
reading config from jdbc string

bojanv55:
Instead of specifying separate config, use spring jdbc url. Supports also dynamic port.


vorburger:
@bojanv55 I'm terribly sorry for reviewing this only today with so much delay!  -- I would welcome code review by other users with more Spring (Boot) experience re. these changes... anyone else willing to help this project by contributing review of this PR?

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
```diff
@@ -19,26 +19,42 @@
...
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.DependsOn;
 
-import javax.sql.DataSource;
-
 //separate with MariaDB4jSpringConfiguration for test of it
 @Configuration
 @ConfigurationProperties("spring.datasource")
+@ConditionalOnProperty(
```
vorburger: anyone with more Spring Experience reading this knows what this does?

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
```diff
@@ -19,26 +19,42 @@
...
+    public DataSourceAutoConfiguration(MariaDB4jSpringService mariaDB4j){
+        this.mariaDB4j = mariaDB4j;
+    }
 
     @Bean
     @DependsOn("mariaDB4j")
-    public DataSource dataSource(DataSourceProperties dataSourceProperties) {
+    public DataSource dataSource(DataSourceProperties dataSourceProperties) throws ManagedProcessException {
+        MariaDBUrl mariaDBUrl = new MariaDBUrl(dataSourceProperties.getUrl());
+        mariaDB4j.getDB().createDB(mariaDBUrl.getDb());
```
vorburger: why do we add `createDB()` here, when before we didn't need this?

vorburger:
I would still welcome code review by other users with more Spring (Boot) experience re. these changes... good ideas & useful? Any risks of regressions for existing users?

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
bojanv55: I use this for unit tests, and it's there so I can take db name from connection string and then create DB in a run-time (since it does not exist when run).

vorburger:
@bojanv55 sorry I have not had the time to maintain this project, but picked it up today.

Would you like to rebase this PR, squash your 31 commits into 1, so that I can re-review it now?

Or are you no longer using MariaDB4j and would like this PR to be closed?

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springboot/autoconfigure/DataSourceAutoConfiguration.java
agostop: > anyone with more Spring Experience reading this knows what this does?

this annotation means, this configuration class will be load when condition is match.
prefix: configuration file must have a config item with this prefix. eg: mariaDB4j.xxxx
name:  if a prefix has been defined, it is applied to compute the full key of each property.  eg: mariaDB4j.enabled
matchIfMissing: to answer will the class be loaded if it(the prefix config item) doesn't exist.

agostop: so much more conflict in this pr when I use rebase. I can't reslove them one by one.

Issue #207, opened 2019-02-11
mariaDB4j-maven-plugin is broken on master


vorburger:
`mvn clean install` on current master fails due to:

```
[INFO] --- maven-invoker-plugin:3.2.0:install (prepare-integration-test) @ mariaDB4j-maven-plugin ---
(...)
[INFO] [INFO] --- mariaDB4j-maven-plugin:2.3.1-SNAPSHOT:stop (post-integration-test) @ start-stop ---
[INFO] [INFO] Stopping MariaDB4j...
[INFO] [INFO] mysqld: 2019-02-11 20:37:37 139736656480000 [Note] /tmp/MariaDB4j/base/bin/mysqld (unknown): Normal shutdown
[INFO] [INFO] mysqld: 
[INFO] [INFO] mysqld: 2019-02-11 20:37:37 139736207316736 [Note] InnoDB: FTS optimize thread exiting.
[INFO] [INFO] Program [/tmp/MariaDB4j/base/bin/mysqld, --no-defaults, --console, --skip-grant-tables, --max_allowed_packet=64M, --basedir=/tmp/MariaDB4j/base, --datadir=/tmp/MariaDB4j/data/45257, --port=45257, --socket=/tmp/MariaDB4j.45257.sock] (in working directory /tmp/MariaDB4j/base) just exited, with value 0
[INFO] [INFO] Successfully destroyed Program [/tmp/MariaDB4j/base/bin/mysqld, --no-defaults, --console, --skip-grant-tables, --max_allowed_packet=64M, --basedir=/tmp/MariaDB4j/base, --datadir=/tmp/MariaDB4j/data/45257, --port=45257, --socket=/tmp/MariaDB4j.45257.sock] (in working directory /tmp/MariaDB4j/base)
[INFO] [INFO] Database stopped.
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [INFO] BUILD SUCCESS
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [INFO] Total time: 14.155 s
[INFO] [INFO] Finished at: 2019-02-11T20:37:39+01:00
[INFO] [INFO] Final Memory: 18M/352M
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [INFO] cleanupOnExit() ShutdownHook quietly deleting temporary DB data directory: /tmp/MariaDB4j/data/45257
[...]

PR #209, opened 2019-02-11
make Travis CI do 'mvn install' instead just 'test' (fixes #208)

### LINKED ISSUE #208: Travis CI is useless because it's broken
#207 illustrates that Travis CI as it's currently configured is of limited value, because it missed to flag up the currently broken mariaDB4j-maven-plugin ...

.. I think this may be because it's doing `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V` && `mvn test -B` .. that's not a good idea - just `mvn install -B -V` would be much better?

vorburger:


vorburger:
eh ben voilà, as per #207 the build is actually broken, Travis just didn't show it before this...

I'll rebase this when #207 is merged, then it should pass, and this can be merged.

vorburger:
merged as 5fe77bdc13ee5f69a4344c1fde7ccc70a2a09fa6

PR #248, opened 2019-04-12
Bug/247 remove spring version

### LINKED ISSUE #247: [Maven] spring.version property needs to be declared even when not using spring
MariaDB4j [uses](https://github.com/vorburger/MariaDB4j/blob/cb1eb92b8804a6e49b120c4274ca8b476ab0c50b/pom.xml#L43) `spring.version` property for `spring-core` dependency.

We use MariaDB4j in a library artefact, that gets BOM-imported into other parts of the application. This fails without redeclaring the `spring.version` property in the final destination POM:
```
[ERROR] Failed to execute goal on project example-impl: Could not resolve dependencies for project com.example:example-impl:jar:1.5.0-SNAPSHOT: Failed to collect dependencies at com.example:example-rest-client:jar:1.7.0-SNAPSHOT -> com.example:common-test:jar:1.15.0-SNAPSHOT -> ch.vorburger.mariaDB4j:mariaDB4j:jar:2.3.0 -> ch.vorburger.mariaDB4j:mariaDB4j-core:jar:2.3.0 -> org.springframework:spring-core:jar:${spring.version}: Failed to read artifact descriptor for org.springframework:spring-core:jar:${spring.version}: Could not transfer artifact org.springframework:spring-core:pom:${spring.version} from/to nexus (http://nexus.example.com:8081/repository/public/): Failed to transfer file: http://nexus.example.com:8081/repository/public/org/springframework/spring-core/$%7Bspring.version%7D/spring-core-$%7Bspring.version%7D.pom. Return code is: 400 , ReasonPhrase:Invalid repository path. -> [Help 1]
```

Maven POM import:
```xml
<type>pom</type>
<scope>import</scope>
```
typically used as a BOM-import pattern does not import Maven properties. `import` imports only management sections of the pom file. 

Redeclaring the version is an ugly workaround, because our application does not depend on spring in any way.
This probably can be fixed by simply not using the property version declarations, and using the `<dependencyManagement>` instead.

mdindoffer:
Remove unused spring.version property and fix IT tests
Closes #247

### DIFF
--- a/mariaDB4j-maven-plugin/src/it/mariadb4j-maven-plugin-test-basic/pom.xml
+++ b/mariaDB4j-maven-plugin/src/it/mariadb4j-maven-plugin-test-basic/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>ch.vorburger.mariaDB4j</groupId>
         <artifactId>mariaDB4j-pom</artifactId>
-        <version>2.4.0-SNAPSHOT</version>
+        <version>2.4.1-SNAPSHOT</version>
         <relativePath>../../pom.xml</relativePath>
     </parent>
     <modelVersion>4.0.0</modelVersion>
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,6 @@
 
 	<properties>
 		 <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		 <spring.version>4.3.9.RELEASE</spring.version>
 		 <springboot.version>2.1.3.RELEASE</springboot.version>
 		 <maven.compiler.source>1.8</maven.compiler.source>
 		 <maven.compiler.target>1.8</maven.compiler.target>


vorburger:
@mdindoffer thanks for contributing, and sorry for the slow review (life got into the way)

I rebased this to see if this fixes #262, if it does, I'll merge it immediately. If it does not, then I'll merge it after someone contributes a fix for #262.

mariaDB4j-maven-plugin/src/it/mariadb4j-maven-plugin-test-basic/pom.xml
```diff
@@ -6,7 +6,7 @@
     <parent>
         <groupId>ch.vorburger.mariaDB4j</groupId>
         <artifactId>mariaDB4j-pom</artifactId>
-        <version>2.4.0-SNAPSHOT</version>
+        <version>2.4.1-SNAPSHOT</version>
```
vorburger: @mdindoffer oh I see it's this (not the removal of the spring.version, below) which fixes #262 ...

PR #267, opened 2019-06-26
Update .travis.yml

vorburger:


vorburger:
Hang on, on re-reviewing this one (from myself), I'm actually no longer sure why I did this, or if this is a good idea... on the contrary, I think this would make testing future PRs such as #235 impossible? I just raised #289 to confirm that - I expect it will fail... if that is so, I'll Abandon this.

What may work and be a much more minor Good Idea is simply #290 though?

Issue #287, opened 2019-08-27
mysql_install_db script takes an inconsistent amount of time to terminate self


Anthoknee:
Hello, 

In the CI process for our app, we are seeing an inconsistency in terms of how long it takes for `mysql_install_db` to terminate itself when using `DB.newEmbeddedDB`. This is becoming an issue as it makes the timed health check flaky. 

I have seen anywhere between 5 seconds and over 2 and a half minutes, as below. 

Is there any known reason for this inconsistency? 

Example log of a longer termination below:

```
2019/08/27 18:16:09.431 INFO [ManagedProcess] [main] [application] [] Thread is now going to wait for this process to terminate itself: Program [/export/content/apps/application/tmp/MariaDB4j/base/scripts/mysql_install_db, --datadir=
/export/content/apps/application/tmp/MariaDB4j/data/9653, --basedir=/export/content/apps/application/tmp/MariaDB4j/base, --no-defaults, --force, --skip-name-resolve] (in working directory /export/content/apps/application/tmp/MariaDB4j/base)
2019/08/27 18:16:09.641 INFO [ManagedProcess] [Exec Stream Pumper] [application] [] mysql_install_db: Installing MariaDB/MySQL system tables in '/export/content/apps/application/tmp/MariaDB4j/data/9653' ...
2019/08/27 18:18:46.946 INFO [ManagedProcess] [Exec Stream Pumper] [application] [] mysql_install_db: OK
```

vorburger:
I don't suppose #285 would help?

If not, you'll have to debug it, and maybe raise a PR with an Improvement?

Please support this project via the Sponsor button.

Anthoknee:
#285 unfortunately does not help. 

I've tried debugging the various SQL statements that are run and have been unable to pinpoint one specific statement that causes this to hang. mysql_system_tables_data does throw an error on insertion where `ssl_cipher`, `x509_issuer`, `x509_subject`, and `authentication_string` do not have defined defaults. Even after updating those, it seems to hang on different parts of the install intermittently. 

I ran the script 400 times with different install directories and found times between 5 and 61 seconds, with an average of about 11.22 seconds on my machine.

My proposed fix would be to:

1. Add defaults to the script that are missing values as per above
2. Have a timeout and number of attempts on the install script that is configurable. If not defined, it will just wait for the exit (current behavior). 

To do this: 
1. Add a `destroyed` flag to `ManagedProcess` (watchdog doesn't seem to set this)
[...]

PR #299, opened 2019-09-03
Attempt to install mysql N times, if not able to complete within a configurable timeout

Anthoknee:
These are the changes that I have described in: #287 

The mysql installation process has introduced some flakiness into our CI environment, as it can get stuck and take minutes to complete. 

After some extensive debugging, I was unable to pinpoint a specific cause. So a fix with a bit more debugging is to allow for the mysql install script to be "restarted" if it is taking too long, as per configs. 

Unless these configs are applied, existing behavior of waiting for exit remains. 

Also, I have a corresponding change to ch.vorburger.exec here:
https://github.com/vorburger/ch.vorburger.exec/pull/28 that this PR relies upon.

### DIFF
--- a/mariaDB4j-core/pom.xml
+++ b/mariaDB4j-core/pom.xml
@@ -18,7 +18,7 @@
         <dependency>
             <groupId>ch.vorburger.exec</groupId>
             <artifactId>exec</artifactId>
-            <version>3.1.1</version>
+            <version>3.1.2</version>
         </dependency>
         <dependency>
             <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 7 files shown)


vorburger:
I'm quite reluctant to accept and merge this contribution... I mean, this is a kind of workaround for a very special problem - mysql_install_db is slow on your CI. (Nobody else ever reported a problem like this before.) Shouldn't you instead figure out the root cause of that problem, on your environment? 

What I would be much more open for would be to make it easy for you to do what you propose here in your own code - just subclass and override, where you need? (If there's something that's not visible, then raise a PR for that.)

What do other past contributors 
but also users of this project reading this think?

Anthoknee:
I can subclass in my code for now if there is reluctance. 

So the only thing I would need for now is to make the method `installPreparation()` (which imo is better named `createMysqlInstallProcess`) `protected`. 

Any opposition to this? If no, I can update this PR to just that change, which will enable me to make the other changes locally.

Anthoknee:
Created #300, which represents this same work.

PR #300, opened 2019-09-05
Rename installPreparation method to createMysqlInstallProcess and expose as protected

Anthoknee:
As per #299 , exposing `installPreparation` (and rename the method to something more appropriate) to allow for subclassing of the `install` method to better suit individual project needs.

### DIFF
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
@@ -7,9 +7,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *      http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
```diff
@@ -99,7 +99,7 @@ public static DB newEmbeddedDB(int port) throws ManagedProcessException {
         return newEmbeddedDB(config.build());
     }
 
-    ManagedProcess installPreparation() throws ManagedProcessException, IOException {
+    protected ManagedProcess createMysqlInstallProcess() throws ManagedProcessException, IOException {
```
vorburger: I'm trying to avoid any MySQL vs MariaDB naming, especially in the API, how about just:

```suggestion
    protected ManagedProcess createDBInstallProcess() throws ManagedProcessException, IOException {
```

Anthoknee: Good call.

vorburger:
@Anthoknee rename the declaration as well, so that the build doesn't fail? ;-)

vorburger:
> @Anthoknee rename the declaration as well, so that the build doesn't fail? ;-)

@Anthoknee were you still planning to do this?

vorburger:
@Anthoknee thanks again for this contribution!  Really really sorry it took so long to merge this.

PR #335, opened 2020-02-21
Added option to run database scripts less verbose

JD-CSTx:
Added an option to run database scripts less verbose. 

This runs fine with all unit tests, as you can see it almost changes nothing in the code!

But: If I create big databases for integration tests, it safes about a minute where it otherwise would scroll seemingly endlessly through the SQL script file content I already know. And without this option my log files are flooded with none-info. I could change the logging level, but all the other info from MariaDB4j is really informative, but not the script file, content I already know re-listed in excruciating detail.

This change is so small that I wouldn't insist on an entry in the contributors.md.

### DIFF
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
@@ -313,7 +313,7 @@ public class DB {
     }
 
     public void run(String command, String username, String password, String dbName) throws ManagedProcessException {
-        run(command, username, password, dbName, false);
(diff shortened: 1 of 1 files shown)


vorburger:
You would be very very welcome to add yourself to contributors.md nevertheless... ;-)

vorburger:
I can't seem to merge this on mobile in the train while I'm en route to the 🎿 ski slopes, so I'll do it later.

JD-CSTx:
Well, I wanted to remind you that you wanted to merge this later.

JD-CSTx:
Thanks 👍

PR #382, opened 2020-10-10
[MariaDB4jService] Do not recreate the DB if already running

Tomlincoln:
I think the start method should check if already started or not.
If you make the MariaDB4jSpringService bean from a config class, and you run source on it, then after initializing all beans, when PostConstruct is called, it calls start() and _resets_ the DB. I think this is not intentional, so I've added the necessary check.


vorburger (approved):
LGTM

FTR: Build is green on https://travis-ci.org/github/vorburger/MariaDB4j/builds/734532843 (no idea what is/was wrong with Travis).

Issue #372, opened 2020-08-28
Add automatic-module name to manifests


jensli:
Add an entry for automatic module name to the JAR manifest files of the runtime libraries.

I get warnings when building my project because the the MariaDB4j module name is derived from the JAR name.

Such a manifest entry looks like this:

Automatic-Module-Name: ch.vorburger.mariadb4j

It can be generated with the Maven.

More information here:

http://branchandbound.net/blog/java/2017/12/automatic-module-name/

MariaDB4j-version: 2.6.2

vorburger:
@jensli you are very welcome to contribute a pull request with what you need to this project!

jensli:
> @jensli you are very welcome to contribute a pull request with what you need to this project!

I spent some time trying to figure out how to do this actually, but I didn't understand to which POM files I should add the configuration.

It is not possible for me to spend more time on it, sorry.

vorburger:
Pending PR #387, pending a new minor ch.vorburger.exec version release.

jensli:
Thanks for a nice tool Michael! I think this change will help it too stay relevant for post-9 Java versions.

Issue #334, opened 2020-02-19
Release next version 2.5.0


egwepas:
Hello, this is a simple question : do you have any plan to release new version with latest fixes (especially the spaces in windows pathes in my case).

In all cases thanks for your helpful tool !

xeester:
Hello, same concern for me : fixing problems with spaces in windows pathes would be very great!
Thank you!

vorburger:
@egwepas @xeester do you want to raise a PR to help fix the broken build? See #451.

vorburger:
OK so I'm piecing from memory where I left this off:

The test failure in #451 is but only blocker which needs to be addressed before a release.

I'm also hitting #452, but that's probably just a local issue for me, which I can separately fix.

For a local build, I first needed to `mvn install` the "new" (cough) `MariaDB4j/DBs/mariaDB4j-db-10.3`. That will have to be deployed, first.

For #372, I then need to `mvn deploy` a new cut of https://github.com/vorburger/ch.vorburger.exec.

Only then as final next step I can deploy a new version of MariaDB4j itself.

But first I have to set up my GPG key to be able to deploy to Maven central again! It's been a while.

valdo404:
Hi, still no release ? How can we help ?

vorburger:
@valdo404 et al reading along here, please [Sponsor my maintenance this project](https://github.com/sponsors/vorburger)! 

It's purely symbolic - but I would appreciate it.

vorburger:
@valdo404 et al reading along here, a 2.5.0 release is currently blocked by #488. Contributions welcome.

vorburger:
Release 2.5.0 has just been pushed to [Maven Central](https://mvnrepository.com/artifact/ch.vorburger.mariaDB4j/mariaDB4j). It should appear on a mirror near you shortly (but it can take at least a few hours).

vorburger:
https://search.maven.org/artifact/ch.vorburger.mariaDB4j/mariaDB4j/2.5.0/jar shows it now.

https://mvnrepository.com/artifact/ch.vorburger.mariaDB4j/mariaDB4j does not show it yet (just about 24h after the release), but that site probably just take a bit longer to update.

vorburger:
https://twitter.com/vorburger/status/1465115301960470531

PR #387, opened 2020-10-14
Issue #372: Add automatic module name to core

jensli:
Libraries without a set automatic module name causes build warnings for
all clients who build with Java >= 9.

Also bump version of the Exec library, to use a version which has the
automatic module name set.

The following pull request for Exec adds an automatic module name:
https://github.com/vorburger/ch.vorburger.exec/pull/40

### DIFF
--- a/mariaDB4j-core/pom.xml
+++ b/mariaDB4j-core/pom.xml
@@ -18,7 +18,7 @@
         <dependency>
             <groupId>ch.vorburger.exec</groupId>
             <artifactId>exec</artifactId>
-            <version>3.1.1</version>
+            <version>3.1.2</version>
         </dependency>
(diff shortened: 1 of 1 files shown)


vorburger:
This will fail to build unless a new https://github.com/vorburger/ch.vorburger.exec v3.1.2 is released.. I'll see if I can get to that some time.

A248:
Hey @vorburger , I don't want to bug you about this, but do you think we could have some progress on this? It's been almost 4 years since Java 9 was released.

Running tests on the module path really requires stable module names. Using something like `requires exec` is highly unstable and will break as soon as another jar called "exec" is introduced to the module path.

vorburger:
See #334.

PR #403, opened 2020-11-11
Adding option to specify os user to the mysqld process

m80592:
When running a service build in drone, I was trying to use mariadb4j as an embedded test database.
It refused to start as drone runs all the processes as root in the container and mysqld complained/refused to start as the user was root.
Found a workaround to specify the os user, to circumvent the problem.
I believe this will be helpful for users who will try to use with springboot in kubernetes environments.


vorburger (approved):
LGTM

PR #476, opened 2021-10-26
mariaDB4j-maven-plugin: exposed database url as Maven Project property

robin-xyzt-ai:
There is currently no way to know on which port the `mariaDB4j-maven-plugin` has started the database so that integration tests can connect to it.
The only option is to hardcode a port number in the configuration of the `mariaDB4j-maven-plugin`, and use that same hardcoded port number in your test.

The old plugin version from mike10004 exposed a port number as Maven Project property (see [this file](https://github.com/mike10004/mariadb4j-maven-plugin/blob/master/mariadb4j-maven-plugin/src/main/java/com/github/mike10004/mariadb4jmavenplugin/Mariadb4jFindFreePortMojo.java)).

This PR does more or less the same. Instead of exposing the port number, it exposes the database URL. I also updated the readme to show how to retrieve that database URL and pass it to your integration tests as system property.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,51 @@ helpful tool for launching MariaDB from Java.
 
 See pom and integration test in https://github.com/vorburger/MariaDB4j/tree/mariaDB4j-maven-plugin/mariaDB4j-maven-plugin/src/it/mariadb4j-maven-plugin-test-basic  for usage example.
 
+#### Example usage
+
+An example usage of this plugin is to install and start a database at the start of the integration test phase, and stop and uninstall the database afterwards.
+
+This is done by configuring the plugin to execute the `start` goal in the `pre-integration-test` phase and the `stop` goal in the `post-integration-test` phase:
+
+```xml
+<plugin>
+  <groupId>ch.vorburger.mariaDB4j</groupId>
+  <artifactId>mariaDB4j-maven-plugin</artifactId>
+  ...
+  <executions>
+    <execution>
+      <id>pre-integration-test</id>
+      <goals>
+        <goal>start</goal>
+      </goals>
+    </execution>
+    <execution>
+      <id>post-integration-test</id>
+      <goals>
+        <goal>stop</goal>
+      </goals>
+    </execution>
+  </executions>
+</plugin>
+```
+
+This will ensure there is a MariaDB instance running on a random port, and expose the database URL as a Maven Project property.
+
+To access the database in your integration tests, you can pass the database URL as system property to your integration tests:
+
+```xml
+<plugin>
+  <groupId>org.apache.maven.plugins</groupId>
+  <artifactId>maven-failsafe-plugin</artifactId>
+  ...
+  <configuration>
+    <systemProperties>
+      <mariadb.databaseurl>${mariadb4j.databaseurl}</mariadb.databaseurl>
+    </systemProperties>
+  </configuration>
+</plugin>
+```
+
 #### How to upgrade the maven plugin from mike10004 version to this version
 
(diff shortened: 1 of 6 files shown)


vorburger (approved):
LGTM! @robin-xyzt-ai thank you for your contribution; please keep it coming if you have others.

vorburger (changes requested):
@robin-xyzt-ai actually... sorry, this does LGTM - BUT it's breaking the tests! I'm getting a NPE `at ch.vorburger.mariadb4j.StartMojo.runWithMavenJvm(StartMojo.java:64)`. If this worked for you when you wrote it, then perhaps it broke when I merged #477 or #478 or #482 or #483 today? If you want to see this merged, could you rebase and fix whatever is causing that NPE?

vorburger:
#489 doesn't seem to help to fix the NPE.

vorburger:
@robin-xyzt-ai BTW it would be cool to have "test coverage" for this - do you want to add what you are describing in the README to the https://github.com/vorburger/MariaDB4j/blob/master/mariaDB4j-maven-plugin/src/it/mariadb4j-maven-plugin-test-basic/pom.xml and then read that system property in https://github.com/vorburger/MariaDB4j/blob/master/mariaDB4j-maven-plugin/src/it/mariadb4j-maven-plugin-test-basic/src/test/java/ch/vorburger/mariadb4j/mariadb4jmavenplugintest/basic/BasicUsageIT.java?

robin-xyzt-ai:
I wanted to write tests for this, but at the time of this PR the build was failing so I couldn't run the tests through Maven. Perhaps I should have set up a real project and do it in there. I will have another look at this

robin-xyzt-ai:
Added a commit which fixes the NPE. Fixed it by switching from `lookupMojo` to `lookupConfiguredMojo` in the tests, based on the comments underneath [this blog post](https://vzurczak.wordpress.com/2014/07/23/write-unit-tests-for-a-maven-plug-in/).

I admit that I don't know if this change has any other unwanted effects. It's my first time writing/testing a Maven plugin and the documentation on testing a Maven plugin is a bit lacking. E.g. those methods don't even have decent Javadoc telling me what the difference is between both.

robin-xyzt-ai:
@vorburger : can you tell me how I can run `BasicUsageIT.java` test from Maven.

Currently I'm using the same as what travis is using (`mvn verify -B -V` in the root directory), but that doesn't seem to trigger that `BasicUsageIT.java` test. I could not find it in the Maven output, and when I tried to simply throw an exception in that test the build is still a success.

That makes it a bit hard to test whether my changes to the test have any effect before I commit them.

vorburger:
> @vorburger : can you tell me how I can run `BasicUsageIT.java` test from Maven.
> 
> Currently I'm using the same as what travis is using (`mvn verify -B -V` in the root directory), but that doesn't seem to trigger that `BasicUsageIT.java` test. I could not find it in the Maven output, and when I tried to simply throw an exception in that test the build is still a success.

Oh! I actually never noticed that myself...

@duttonw who originally contributed this in #135 may want to chime in here?

You could also "manually" test it - I wouldn't be against merging this as-is, if you want me to.

robin-xyzt-ai:
I think I've done everything you asked for this PR:

- The NullPointerExceptions in the test are fixed
- Rebased on the current version of master for easier merging
- Added an assertion to the BasicUsageIT (but since I don't know how to run it, I couldn't verify it actually works)

vorburger (approved):

robin-xyzt-ai:
Thank you for accepting and merging this PR.

PR #502, opened 2021-11-30
When dumping the database, close the outputstream once the dump is done

### LINKED ISSUE #340: Output file resources not closed when using DB.dump
Hello,

we are having a problem when using this excellent library to generate a dump of the local data with the provided wrapper to mysqldump (DB.dump).

We need to manipulate the exported data and afterwards, delete the file. However this is not possible in Windows (in linux it works ok), due to the fact that windows doesn't allow to delete opened files. Even using deleteOnExit fails to delete the file (it works ok on other created files).

I suspect that inside DB.dump, a BufferedOutputStream gets created but never really closed, and this causes delete/deleteOnExit to fail, as when it is run the file is still "open"/(not released).

Inside DB.dump:
`builder.addStdOut(new BufferedOutputStream(new FileOutputStream(outputFile)));`

This is a sample of how we use the library to dump data and how we try to delete the file, the file remains open on program exit (afterwards, it can be deleted manually without problem)

```kotlin
val outputFile = File("filepath.sql")
val dumpProcess = db.dumpSQL(outputFile, dbName, user, password)
dumpProcess.start()
dumpProcess.waitForExit()
outputFile.deleteOnExit()
// Manipulate outputFile etc
```

robin-xyzt-ai:
This should fix #340 .

The created outputstream is now closed using a process listener, which should get triggered when the dump is either done or has failed.

Alas, no test in this commit.

- I couldn't find any existing tests that perform a dump
- Even if there would be such a test, I would not know how to check from Java code whether that file stream is closed, unless I added some ugly code to grant me access to that stream

If you really want to have a test for this, some pointers in how you would approach testing this are welcome

### DIFF
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
@@ -19,8 +19,11 @@
  */
 package ch.vorburger.mariadb4j;
 
-import ch.vorburger.exec.*;
-
+import ch.vorburger.exec.ManagedProcess;
+import ch.vorburger.exec.ManagedProcessBuilder;
+import ch.vorburger.exec.ManagedProcessException;
+import ch.vorburger.exec.ManagedProcessListener;
+import ch.vorburger.exec.OutputStreamLogDispatcher;
 import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
@@ -516,10 +519,11 @@ public class DB {
                 closeOutputStream();
             }
 
-            private void closeOutputStream(){
+            private void closeOutputStream() {
                 try {
                     outputStream.close();
-                } catch (IOException ignore) {
+                } catch (IOException exception) {
+                    logger.error("Problem while trying to close the stream to the file containing the DB dump", exception);
                 }
             }
         });
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
@@ -19,10 +19,8 @@
  */
 package ch.vorburger.mariadb4j;
 
-import ch.vorburger.exec.ManagedProcess;
-import ch.vorburger.exec.ManagedProcessBuilder;
-import ch.vorburger.exec.ManagedProcessException;
-import ch.vorburger.exec.OutputStreamLogDispatcher;
+import ch.vorburger.exec.*;
+
 import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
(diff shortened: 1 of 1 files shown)


mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
```diff
@@ -506,6 +508,25 @@ protected ManagedProcess dump(File outputFile, List<String> dbNamesToDump, boole
...
+            }
+
+            @Override
+            public void onProcessFailed(int i, Throwable throwable) {
+                closeOutputStream();
+            }
+
+            private void closeOutputStream() {
+                try {
+                    outputStream.close();
```
vorburger, line 524: @robin-xyzt-ai first of all a big Thank You for stepping up to fix this issue!

But I'm not sure the fix as-is is entirely correct. This line will close the `BufferedOutputStream`, and as far as I know that would not automagically also close that `FileOutputStream` created in line 485? Do you see what I mean and agree to amend this close both of them, or am I wrong and confused?

vorburger (changes requested):

vorburger:
I'm OK to merge this PR for that particular bug without test coverage. The effort to create one far outweighs the benefit in this case, I agree. If you really had and wanted to, you would have to somehow intercept / mock the FS impl... something like https://github.com/google/jimfs comes to mind, but I'm not even sure that has any way to "check if the virtual (in-memory) file has been closed" - probably not. Another route, more fruitful, would probably be to use some platform speicific native library, or even `exec` (LOL, the irony, as a test in a library that is itself an exec wrapper) out to some `lsof` type external binary, to check if the dump file is still open. Long story short - forget about it.

vorburger:
> I couldn't find any existing tests that perform a dump

just FYI there IS a https://github.com/vorburger/MariaDB4j/blob/master/mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleDumpTest.java (note that most of the "interesting" tests are in the mariaDB4j module, not the mariaDB4j-core module where the code is)

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
vorburger: It seems I'm wrong about this... `BufferedOutputStream`, in it's parent `FilterOutputStream.close()` actually does `out.close();` and its JavaDoc explicitly says _"...  pass all requests to the underlying output stream"._

vorburger (approved):

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
robin-xyzt-ai: Indeed. As such, it's not an implementation detail but documented behavior. As such, we could rely on it

Issue #501, opened 2021-11-30
Fix for spaces in data path doesn't work on windows


Blanco27:
Trying to create an embedded DB like this:

```
        final var db = DB.newEmbeddedDB( DBConfigurationBuilder.newBuilder()
            .setDeletingTemporaryBaseAndDataDirsOnShutdown( false )
            .setDataDir( "D:\\contains spaces" )
            .build() );
```

creates two folders:

![image](https://user-images.githubusercontent.com/56546322/144069331-c0fc3a9d-b86a-4225-81f9-fff874deca06.png)

The one that ends up being used is the one with the ``%20` for spaces, the other one is left empty.

If you append another folder behind the one with spaces, you get an exception instead:

```
15:50:51,468 ERROR LoggingExecuteResultHandler:55 - Program [C:\Users\nwe\AppData\Local\Temp\MariaDB4j\base\bin\mysql_install_db.exe, --datadir=D:\contains%20spaces\data] (in working directory C:\Users\nwe\AppData\Local\Temp\MariaDB4j\base) failed unexpectedly
org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
	at org.apache.commons.exec.DefaultExecutor.executeInternal(DefaultExecutor.java:404)
	at org.apache.commons.exec.DefaultExecutor.access$200(DefaultExecutor.java:48)
	at org.apache.commons.exec.DefaultExecutor$1.run(DefaultExecutor.java:200)
	at java.base/java.lang.Thread.run(Thread.java:829)
15:50:51,520 ERROR ManagedProcess:308 - Program [C:\Users\nwe\AppData\Local\Temp\MariaDB4j\base\bin\mysql_install_db.exe, --datadir=D:\contains%20spaces\data] (in working directory C:\Users\nwe\AppData\Local\Temp\MariaDB4j\base) failed
15:50:51,544 ERROR SpringBootstrap:1000 - Konnte APM Server Bootstrap nicht starten.
ch.vorburger.exec.ManagedProcessException: An error occurred while installing the database
	at ch.vorburger.mariadb4j.DB.install(DB.java:143)
	at ch.vorburger.mariadb4j.DB.newEmbeddedDB(DB.java:83)
	at com.isp.lea.service.ioc.SpringBootstrap.startServer(SpringBootstrap.java:684)
	at com.isp.lea.service.ioc.SpringBootstrap.main(SpringBootstrap.java:230)
[...]

PR #505, opened 2021-12-02
Replace DB's toWindowsPath() with inline File.getCanonicalFile()

vorburger:
Because following the previous commit, it's no longer Windows specific.

This is backwards compatible because toWindowsPath() was private.

See https://github.com/vorburger/MariaDB4j/issues/501


vorburger:
@Blanco27 OK for you?

FYI this is just a minor clean-up follow-up to your #504, and not directly related to the https://github.com/vorburger/ch.vorburger.exec/pull/58; I have a separate PR about that coming up.

vorburger:
@Blanco27 Could you double check that this does not break on Windows?

Blanco27:
Still works on Windows 👍

PR #515, opened 2021-12-16
Update pom.xml

dev-fringe:


vorburger:
@dev-fringe thanks! #513 and #514 also do this automated, but I've merged yours instead of those because you change all 3 and dependabot seems to somehow have missed one.

PR #533, opened 2022-01-20
Support default character set configuration.

agostop:
Hi, I add default character set option in configuration. 
I know the addArgument function, but this action only effect with server side. 

When you want run script as another character set, "addArgument" function will doesn't work.

### DIFF
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
@@ -195,6 +195,9 @@ public class DB {
         for (String arg : configuration.getArgs()) {
             builder.addArgument(arg);
         }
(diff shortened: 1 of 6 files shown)


vorburger:
@agostop this looks pretty useful, and I'm happy to merge it. However, I've just noticed that the CI build doesn't work anymore. Please "hold the line" (I mean like when you have to wait on the phone 😄 ) while I sort this out in #538. When there is a working build again, if this passes it I'll merge this PR. Please don't hesitate to ping me again in let's say 1 week in case I should forget.

vorburger:
@agostop actually would you mind to `git rebase` this PR? (I'm assuming you know what that means / how that works, please do feel free to just ask if that's not clear.) The reason is that Travis CI is not re-running the build without a kick,like a re-base.

agostop:
just like this? @vorburger

Issue #602, opened 2022-07-29
Support parallel execution


simonzkl:
So I've been testing MariaDB4j with parallel unit test execution and discovered that in its current form, it does not work. The main issue is that MariaDB instances share the same system-wide `/tmp` directory for temporary tables, and the name of these files often clash between instances, leading to errors like this:
```
Message    : Can't find file: '/tmp/#sql16803_8_ef.MAI' (errno: 2 "No such file or directory")
```

Most often these error occur when running database migrations or when setting up system tables with `mysql_install_db`. The issues with migrations can be worked around by configuring mysqld do run with its own tmpdir (e.g. `--tmpdir=/tmp/m4j-<random token>`. However with the current API, we cannot set the tmpdir for `mysql_install_db`. I've patched this in my code by writing a wrapper around `mariadb4j.DB` and overriding `createDBInstallProcess`. That seems to have solved the issue for me and now tests can execute in parallel without any issues.

My suggestion for mariadb4j would be to configure tmpdir isolation by default for both `mysqld` and `mysql_install_db`. We could probably namespace the data dir and sockets in the same base directory. I can prepare a PR for this, just wanted to confirm if you agree with this approach.

vorburger:
That sounds like a great idea and plan - PR for this most welcome!

vorburger:
@simonzkl with #604 (just now) merged, can we close this issue, or did you want to make further contributions for this?

simonzkl:
Nope, that tmpdir isolation was all that was needed here. I think it's fine to close this. Thanks!

PR #604, opened 2022-08-01
Add configurable tmpdir

simonzkl:
This makes `--tmpdir` configurable, defaulting to `$TMPDIR/MariaDB4j/tmp/$PORT`. It should solve the issue of running multiple `mysqld` instances in parallel. See #602 for more details.

### DIFF
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBShutdownHook.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBShutdownHook.java
@@ -66,16 +66,18 @@ class DBShutdownHook extends Thread implements FileVisitor<Path> {
     private final Supplier<ManagedProcess> mysqldProcessSupplier;
     private final Supplier<File> dataDirSupplier;
     private final Supplier<File> baseDirSupplier;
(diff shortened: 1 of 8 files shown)


mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBShutdownHook.java
```diff
@@ -283,8 +285,6 @@ private boolean isEmptyDirectory(Path directory) throws IOException {
 
     @Override public void run() {
         ManagedProcess mysqldProcess = mysqldProcessSupplier.get();
-        File dataDir = dataDirSupplier.get();
-        File baseDir = baseDirSupplier.get();
```
simonzkl, line 287: I moved these because the linter was complaining they were too far away from where they're used after adding `tmpDir`.

vorburger: I see, thanks for the explanation; that SGTM!

vorburger (approved):
@simonzkl Thank You for contributing this! Sorry it took me a while to code review and merge it.

I welcome further contributions from users!

Issue #497, opened 2021-11-29
Mac M1 binaries


vorburger:
For MariaDB4j to "just" work on M1 Mac... either:

A. Someone contributes binaries for MariaDB binaries JAR for M1

-OR-

B. #296 is implemented to support running MariaDB in a container instead of from binaries in a JAR

I welcome community contributions for A. (but won't work on that myself), but think B. is a better way to pursue for the future.

vorburger:
https://github.com/docker/for-mac/issues/5389 has something about (now solved?) performance issues on M1.

https://stackoverflow.com/questions/69896059/is-it-possible-to-get-mariadb4j-to-work-on-an-m1-mac asks for this.

vorburger:
Just like for #298 re. ARM, https://downloads.mariadb.org doesn't actually seem to have DB binaries for M1, but presumably someone interested in supporting this knows of some place where prebuilt binaries for Mac M1 arch can be downloaded from.

balamsoft:
Well, there's a really easy way to build MariaDB on Mac using Homebrew.
`brew install mariadb --build-from-source` 

As of now this command will build MariaDB 10.6.4.

balamsoft:
I have built it using an M1 Mac Mini and I have the binaries.
```
➜  server git:(109fc67d4da) brew install mariadb --build-from-source

==> Downloading https://ghcr.io/v2/homebrew/core/groonga/manifests/11.0.9
######################################################################## 100.0%
==> Downloading https://ghcr.io/v2/homebrew/core/groonga/blobs/sha256:891b7a85bf8e539d276f3251648c471ce925e401f99b543316ad3c39d794b801
==> Downloading from https://pkg-containers.githubusercontent.com/ghcr1/blobs/sha256:891b7a85bf8e539d276f3251648c471ce925e401f99b543316ad3c39d794b801?
######################################################################## 100.0%
==> Downloading https://downloads.mariadb.com/MariaDB/mariadb-10.6.4/source/mariadb-10.6.4.tar.gz
######################################################################## 100.0%
==> Installing dependencies for mariadb: groonga
==> Installing mariadb dependency: groonga
==> Pouring groonga--11.0.9.arm64_monterey.bottle.tar.gz
🍺  /opt/homebrew/Cellar/groonga/11.0.9: 940 files, 51.3MB
==> Installing mariadb
==> Patching
[...]

PR #628, opened 2022-10-17
Add support for macos-arm64 platform

stevenodb:
- Added support for the macos-arm64 platform
- Unfortunately, no lead right now on a reputable source for an apple silicon mariadb build


vorburger:
> * Unfortunately, no lead right now on a reputable source for an apple silicon mariadb build

Which makes this a No Go to merge (until that's solved), unfortunately. I've commented in #497 about possible next step ideas.

PR #638, opened 2022-11-10
Add workflow for a maven verify on windows

jejanim:


jejanim:
@vorburger I added a **very basic** workflow.
However, it seems that not only the test for the `-springboot` module fails... 
I had to fix another test that uses hard-coded path separators that obviously work on linux only (`/` vs `\`).

Finally, I hit a wall with the following error:
_(see pipeline output here: https://github.com/jejanim/MariaDB4j/actions/runs/3436523459/jobs/5730135732)_
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.4.1:jar (attach-javadocs) on project mariaDB4j-core: MavenReportException: Error while generating Javadoc: 
Error: [ERROR] Exit code: 1 - error: --add-modules ALL-MODULE-PATH can only be used when compiling the unnamed module
Error:  
Error:  Command line was: cmd.exe /X /C "C:\hostedtoolcache\windows\Java_Adopt_jdk\11.0.16-101\x64\bin\javadoc.exe @options @packages @argfile"
```

.github/workflows/buildTestWindows.yml
```diff
@@ -0,0 +1,26 @@
+name: Java CI
+
+on: [push]
```
vorburger, line 3: Let's make this run for all PRs (including this one!), not just AFTER I merge, so [like here] (https://github.com/vorburger/MariaDB4j/blob/main/.github/workflows/codeql-analysis.yml), use:

```suggestion
on:
  push:
    branches: [ main ]
  pull_request:
    # The branches below must be a subset of the branches above
    branches: [ main ]
```

.github/workflows/buildTestWindows.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  test:
```
vorburger, line 6:
```suggestion
  testOnWindows:
```

vorburger (changes requested):

.github/workflows/buildTestWindows.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK 11
+        uses: actions/setup-java@v3
+        with:
+          java-version: '11'
+          distribution: 'adopt'
+          cache: maven
+      - name: Build with Maven
+        run: mvn --batch-mode --update-snapshots verify
```
vorburger, line 23:
```suggestion
        run: mvn --show-version --batch-mode --strict-checksums verify
```

Adding `--show-version` will let us see what Maven version runs here, to understand if that JavaDoc error is related to that (on Travis it's Apache Maven 3.8.6).

Removing `--update-snapshots` should't cause any issues, but be faster.

Adding `--strict-checksums` seems like a good idea, security wise.

I'm making [the Travis CI configuration](https://github.com/vorburger/MariaDB4j/blob/main/.travis.yml) similar with https://github.com/vorburger/MariaDB4j/commit/aeda8bea13e0f932e05a42bf24391e2b208cc9c4.

vorburger (changes requested):

vorburger:
> ```
> Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.4.1:jar (attach-javadocs) on project mariaDB4j-core: MavenReportException: Error while generating Javadoc: 
> Error: [ERROR] Exit code: 1 - error: --add-modules ALL-MODULE-PATH can only be used when compiling the unnamed module
> Error:  
> Error:  Command line was: cmd.exe /X /C "C:\hostedtoolcache\windows\Java_Adopt_jdk\11.0.16-101\x64\bin\javadoc.exe @options @packages @argfile"
> ```

We (me, or anyone else reading along) can now see this problem as well on the (new) Windows Builder Action's Build Log.

It's strange that this doesn't appear on the Linux build on Travis CI. I suspect that perhaps this isn't related to Windows VS Linux, but could be something else... what else than the OS is different here?

The Adopt vs OpenJDK distro difference? Unlikely, I think; but I'm not 100% sure, of course. We COULD play with other https://github.com/actions/setup-java#supported-distributions but I have a hunch that it's not that.

The `maven-javadoc-plugin` version is the same, of course, as it's fixed by the POM; it's v3.4.1 for both.

Maven version difference? Maybe more likely, let's compare that first; I've just made a review suggestion to add `--show-version` to let us see what Maven version actually runs here, to understand if that JavaDoc error is related to that. On Travis where it's works Apache Maven 3.8.6.

Can you accept the proposed changes so that we get a new build and can compare the Maven version? If you see it being older than the 3.8.6, which is the latest on https://maven.apache.org/download.cgi, perhaps you would like to look into how this can remove the older one (if it is) and fetch and use that 3.8.6 version, instead of whichever one comes with the base image that you use here?

jejanim:
Will look into it.

chrisbloe:
Hello! (has been a while ;) - thanks for your continued efforts)

It seems the problem is with Java 11 - https://issues.apache.org/jira/browse/MJAVADOC-586

Could we try building against Java 17? I realise this could open a can of worms!

jejanim:
> Hello! (has been a while ;) - thanks for your continued efforts)
> 
> It seems the problem is with Java 11 - https://issues.apache.org/jira/browse/MJAVADOC-586
> 
> Could we try building against Java 17? I realise this could open a can of worms!

Sure, let's give it a try.

vorburger:
> The one thing that remains is the test failure we wanted.

#647 by @MazenAmria takes a stab at fixing that - but somehow doesn't quite work, yet; see #649. 

I would ideally like to merge both this and #647 together, once they fully pass a build (together).

> Could we try building against Java 17? I realise this could open a can of worms!

@jejanim @chrisbloe This in effect opens the backdoor for this library to require Java 17 instead of only Java 11.  I'm slightly reluctant to "force" this upon all users as part of a PR which adds test coverage on Windows. Do you think it may be possible to fix or workaround that Maven Javadoc Plugin problem differently? I've also opened #650 to get some "signal" (feeling) for how "the community" may feel about this.

vorburger:
Sorry for the long silence on this. I've fixed a number of others things in this project today, and am hoping to find time another weekend to upgrade the project overall from Java 11 to Java 17 and then be able to pick this PR up....

PS: If you use MariaDB4j professionally, please consider either taking a https://tidelift.com subscription or GitHub Sponsor to support future maintenance!

vorburger:
@jejanim I have FINALLY (just now) merged your work here as part of #649! Therefore closing this PR now.

I have squashed your commits here into 1 single https://github.com/vorburger/MariaDB4j/pull/649/commits/7b87da90e3f10c49e9c89c608a7f8947b7a9885a (with you still credited as the Author of it).

Many many thanks again for this contribution!

PR #647, opened 2022-12-06
Removing `--tmpdir` argument from `mysql_install_db` as it's used in `mysqld`

MazenAmria:
Resolving https://github.com/vorburger/MariaDB4j/issues/632 and https://github.com/vorburger/MariaDB4j/issues/634

PR #649, opened 2022-12-10
Test PR to see if the Windows issue is fixed

vorburger:
This PR combines #647 with #638 to see if it works (together).


vorburger:
Huh - interesting... this seems to fail differently, and even earlier already, than #638 did.

@MazenAmria @jejanim would you be interested in further analyzing what's going on here?

MazenAmria:
@vorburger sure

MazenAmria:
@vorburger 

https://github.com/vorburger/MariaDB4j/blob/b71a54494dc01e6d00cff7818a632e3a12fea122/mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleTutorialTest.java#L50-L58

This is trying to use the path `/usr` as a base directory which doesn't exist in Windows.
The test is trying to run `D:\usr\scripts\mysql_install_db.exe`, note that other tests was running something like:`C:\Users\RUNNER~1\AppData\Local\Temp\MariaDB4j\base\bin\mysql_install_db.exe`.

btw, the test that was failing due to `--tmpdir` has passed, I'm not sure if that's enough to verify my PR.

vorburger:
> This is trying to use the path `/usr` as a base directory which doesn't exist in Windows. The test is trying to run `D:\usr\scripts\mysql_install_db.exe`, note that other tests was running something like:`C:\Users\RUNNER~1\AppData\Local\Temp\MariaDB4j\base\bin\mysql_install_db.exe`.

Oh! Right; well spotted. So that test covers the https://github.com/vorburger/MariaDB4j#how-using-existing-native-mariadb-binaries scenario. @jejanim do you think it would be possible to, like [for the Linux tests](https://github.com/vorburger/MariaDB4j/blob/main/.travis.yml#L19), install MariaDB4j on Windows in the Github Workflow? (Somehow; I don't know how.) 

Then we could have this test pass, and cover that use case on Windows (which would be nice...) by putting an `if (config.isWindows())` (by refactoring to pull the `config.build()` out/up) and setting the base dir and executable to wherever it typically is installed on Windows.

vorburger:
> possible to install MariaDB4j on Windows in the Github Workflow? (Somehow; I don't know how.)

https://github.com/ankane/setup-mariadb from @ankane looks pretty interesting for this!

vorburger:
Sorry for the long silence on this. I've fixed a number of others things in this project today, and am hoping to find time another weekend to upgrade the project overall from Java 11 to Java 17 and then be able to pick this PR up....

PS: If you use MariaDB4j professionally, please consider either taking a [Tidelift subscription](https://tidelift.com) or [GitHub Sponsor](https://github.com/sponsors/vorburger) to support future maintenance!

Issue #699, opened 2023-03-11
Add latest MariaDB v10.6+ (incl. support for OpenSSL 3)


cverges-medallia:
OpenSSL 3.x is now out, but MariaDB4j seems to still be using versions that rely on libssl.1.0.0.  (#278 for example.)  Downgrading to OpenSSL 1.0 on newer OS installations causes other system problems that make such a workaround infeasible.

Newer versions of MariaDB (10.6+) appear to have OpenSSL 3.x support (https://jira.mariadb.org/browse/MDEV-25785).  If possible, can the base version of MariaDB provided for by MariaDB4j be upgraded?

vorburger:
@cverges-medallia thanks for your interest in this project!

An upgrade is certainly possible, but I would need a volunteer contributor (you?) to do the actual work to raise a PR...

https://github.com/vorburger/MariaDB4j#mariadb-database-jars-and-version-upgrades actually describes how to!

FYI multi platform support has historically been a PITA. I'll shortly be adding Windows coverage in CI (watch #649), and macOS M1 is a PITA (see #497 and #628). If you want to contrib to the project and raise PR, I would be happy to review and merge it if OK.

**PS: If you use MariaDB4j professionally, please consider to, either:**
* [Taking a Tidelift subscription](https://tidelift.com/), or
* [GitHub Sponsor me](https://github.com/sponsors/vorburger),

**to support future maintenance!**

cverges-medallia:
Are there any instructions that the prior contributors used to allow for reproduction of the binaries that are included in `DBs`?  Without those, the effort is significantly more.

vorburger:
Did you notice that there are the "old" (initial, original) binary ones, and the "newer scripted from source" ones. Ignore the former, and look only at the latter, [e.g. this one](https://github.com/vorburger/MariaDB4j/blob/main/DBs/mariaDB4j-db-10.3/mariaDB4j-db-linux64-10.3/prepare.xml) . Does that help? PS: A first PR with README doc update to clarify this would be great!

cverges-medallia:
Unless we attempt to compile MariaDB statically (which I've been trying to do but seems VERY difficult given the cmake build system), it isn't worth trying to fix generically given how problematic Mac and all the MariaDB dependencies are.  I've opted to just provide my own autoconfig classes with mechanisms for detecting Mac and Homebrew paths appropriately whilst simultaneously still allowing for Linux and Windows support to be transparent.

[...]

Issue #344, opened 2020-03-31
MariaDB 10.6+ binaries packaging


cobbce:
Wanting to get my hands on that sweet sweet Insert... Returning capability and save myself an extra db round trip.

Very excited to discover this project, I'd left H2 behind long since as they didn't keep up with mysql/mariadb syntax evolution, forcing me to write conditionals in production code to change up my syntax, which kind of defeats the purpose of writing integration tests (because the production code path is no longer exercised).

Basically this project is wonderful, keep up the good work!

vorburger:
Thank You! Your contribution to add 10.5 would be very very welcome. There is doc here https://github.com/vorburger/MariaDB4j/blob/master/README.md#mariadb-database-jars-and-version-upgrades how to do it. Just ask if not clear, and we'll try to help.

benweizhu:
We need the 10.5+ too.

luigiberrettini:
Hi @vorburger I read the README and saw that binaries are not supposed to be added to this repo anymore.

Besides the changes to POMs like [the ones you did](https://github.com/vorburger/MariaDB4j/commit/7fb1bd9c625f0867a425a23ba554e2095f919cda) recently, I am not sure where contributors need to submit a PR with the binaries, for them to be packaged as JAR and pushed to Maven central.

vorburger:
>  I am not sure where contributors need to submit a PR with the binaries, for them to be packaged as JAR and pushed to Maven central.

Here in this project, but you need to raise a PR like #235 - without binaries, but just downloading from a safe well known URL.

> read the README and saw that binaries are not supposed to be added to this repo anymore.

I was trying to find where that's written clearly, and myself can't see it now.. where did you see this? I've tried to point someone else to it over in #363. If you want to contribute to the project, a PR updating the README to make it clearer how to do this for contributors in the future would be very welcome.

luigiberrettini:
When talking about binaries I was referring to the DB binaries and as per some issues (not the README) and what you said above they should no more be committed to git and therefore they will be no more available in this repo.

The new way to provide MariaDB updated versions is, per my understanding, creating folders and pom.xml inside them as for the 10.3 version: binaries will be downloaded from known locations on build and used to create JARs but not be anymore stored in this repo.

I would suggest keeping one folder per major.minor version and trying to update the patch version as much as possible.

luigiberrettini:
My understanding is that Travis CI is used to build the DB JARs and the MariaDB4j JAR.

[...]

PR #710, opened 2023-03-28
Initial working pass of MariaDB 10.6.12...

TheKnowles:
Initial working pass of MariaDB 10.6.12 for linux with a simulated start for winx64. OSX support is currently not available.  

Opening this PR knowing it cannot currently be merged to main. I am looking for additional eyes on this and suggestions.

Originally, I branched off of the last commit prior to the upgrade to Java 17. 

After I attempted to use it in my software (Java 11, Spring 5 / Boot 2.7.x), I ran into an issue where the annotations had been switched from javax to jakarta as part of the move to Spring 6. I then moved my branch to right before this change.

I would like to suggest creating a 2.x release branch that maintains a Java 11 / Spring 5 release and allow 3.x + to be Java 17 onwards. But this becomes a nightmare quickly because more recent MariaDB instances are not necessarily tied to the user's Java version, nor what version of Spring they are using. Fixes to MariaDB4j itself would then need to get backpatched to prior release branches as applicable.

The compatibility matrix effectively becomes MariaDB / Spring Framework / this product.

I don't mind doing the work for this, but at what point do we want cutoff retrofitting here versus maintaining a hard fork? Open to suggestions.

Having that said, this update was quite bespoke. In no particular order, several challenges that had to be solved to get 10.6.12 working in Linux:

1. MariaDB changed their download urls since the 10.3 commit to this repository.
2. MariaDB 10.4 removed no password root user. Additional commentary below
3. MariaDB 10.5.2 onward swapped how they did sym links between mysql* and maria* equivalents. Additional commentary below
4. OSX support was incomprehensible.
5. Consolidated slf4j to version 2.0.6. 1.7.x and 2.x results in NOOP binding

**DL URL**

https://dlm.mariadb.com/browse/mariadb_server/ ... versus the previous url in 10.3

Looks like Win32 support was dropped at some point.

**Root user:**

https://mariadb.com/kb/en/authentication-from-mariadb-104/

You can use the user that owns the data directory (generally the user that started MariaDB4j) to give the root user a password to continue the functionality of the unit tests.

**Sym links:**

I updated the entire codebase to reference maria* instead of mysql for binaries.

**OSX:**

Bintray was decommisioned in May 2021. Homebrew OCI got moved to Github Content Repository which is a Docker repo. After a lot of hunting and trying to figure out how to at least simulate a download, I got stuck at pulling the OCI from GHCR here: 

https://github.com/Homebrew/homebrew-core/pkgs/container/core%2Fmariadb%2F10.6/69479903?tag=10.6.12

Because I was not on the right OS, docker would not pull. 

I believe people have ran into this before and effectively have to build from source themselves or be on osx.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ MariaDB4j JAR binaries are available from:
 <dependency>
     <groupId>ch.vorburger.mariaDB4j</groupId>
     <artifactId>mariaDB4j</artifactId>
-    <version>3.0.0</version>
+    <version>2.7.0</version>
 </dependency>
 ```
 
@@ -146,7 +146,7 @@ source; just git clone this and then mvn install or deploy. -- MariaDB4j's Maven
 <dependency>
     <groupId>ch.vorburger.mariaDB4j</groupId>
     <artifactId>mariaDB4j</artifactId>
-    <version>3.0.0-SNAPSHOT</version>
+    <version>2.7.0-SNAPSHOT</version>
 </dependency>
 ```
 
--- a/mariaDB4j-app/pom.xml
+++ b/mariaDB4j-app/pom.xml
@@ -6,7 +6,7 @@
 	<parent>
 		<groupId>ch.vorburger.mariaDB4j</groupId>
 		<artifactId>mariaDB4j-pom</artifactId>
-		<version>3.0.0-SNAPSHOT</version>
+		<version>2.7.0-SNAPSHOT</version>
 		<relativePath>../pom.xml</relativePath>
 	</parent>
 
--- a/mariaDB4j-core/pom.xml
+++ b/mariaDB4j-core/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>ch.vorburger.mariaDB4j</groupId>
         <artifactId>mariaDB4j-pom</artifactId>
-        <version>3.0.0-SNAPSHOT</version>
+        <version>2.7.0-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
 
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
@@ -109,7 +109,7 @@ public class DB {
         logger.info("Installing a new embedded database to: " + baseDir);
         File installDbCmdFile = configuration.getExecutable(Executable.InstallDB);
         ManagedProcessBuilder builder = new ManagedProcessBuilder(installDbCmdFile);
-        builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mysql_install_db"));
+        builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mariadb-install-db"));
         builder.getEnvironment().put(configuration.getOSLibraryEnvironmentVarName(), libDir.getAbsolutePath());
         builder.setWorkingDirectory(baseDir);
         if (!configuration.isWindows()) {
@@ -155,7 +155,7 @@ public class DB {
             mysqldProcess = startPreparation();
             ready = mysqldProcess.startAndWaitForConsoleMessageMaxMs(getReadyForConnectionsTag(), dbStartMaxWaitInMS);
         } catch (Exception e) {
-            logger.error("failed to start mysqld", e);
+            logger.error("failed to start mariadbd", e);
             throw new ManagedProcessException("An error occurred while starting the database", e);
         }
         if (!ready) {
@@ -174,7 +174,7 @@ public class DB {
 
     synchronized ManagedProcess startPreparation() throws ManagedProcessException, IOException {
         ManagedProcessBuilder builder = new ManagedProcessBuilder(configuration.getExecutable(Server));
-        builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mysqld"));
+        builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mariadbd"));
         builder.getEnvironment().put(configuration.getOSLibraryEnvironmentVarName(), libDir.getAbsolutePath());
         builder.addArgument("--no-defaults"); // *** THIS MUST COME FIRST ***
         builder.addArgument("--console");
@@ -203,7 +203,7 @@ public class DB {
         // because cleanupOnExit() just installed our (class DB) own
         // Shutdown hook, we don't need the one from ManagedProcess:
         builder.setDestroyOnShutdown(false);
-        logger.info("mysqld executable: " + builder.getExecutable());
+        logger.info("mariadbd executable: " + builder.getExecutable());
         return builder.build();
     }
 
@@ -342,7 +342,7 @@ public class DB {
         logger.info("Running a " + logInfoText);
         try {
             ManagedProcessBuilder builder = new ManagedProcessBuilder(configuration.getExecutable(Client));
-            builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mysql"));
+            builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mariadb"));
             builder.setWorkingDirectory(baseDir);
             builder.addArgument("--default-character-set=utf8");
             if (username != null && !username.isEmpty()) {
@@ -485,7 +485,7 @@ public class DB {
 
         BufferedOutputStream outputStream = new BufferedOutputStream(new FileOutputStream(outputFile));
         builder.addStdOut(outputStream);
-        builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mysqldump"));
+        builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mariadb-dump"));
         builder.addArgument("--port=" + configuration.getPort());
         if (!configuration.isWindows()) {
             builder.addFileArgument("--socket", getAbsoluteSocketFile());
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfigurationBuilder.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfigurationBuilder.java
@@ -43,7 +43,7 @@ import org.apache.commons.lang3.SystemUtils;
  */
 public class DBConfigurationBuilder {
 
-    protected static final String WIN32 = "win32";
+    protected static final String WINX64 = "winx64";
     protected static final String LINUX = "linux";
     protected static final String OSX = "osx";
 
@@ -54,7 +54,7 @@ public class DBConfigurationBuilder {
     private String databaseVersion = null;
 
     // all these are just some defaults
-    protected String osDirectoryName = SystemUtils.IS_OS_WINDOWS ? WIN32 : SystemUtils.IS_OS_MAC ? OSX : LINUX;
+    protected String osDirectoryName = SystemUtils.IS_OS_WINDOWS ? WINX64 : SystemUtils.IS_OS_MAC ? OSX : LINUX;
     protected String baseDir = SystemUtils.JAVA_IO_TMPDIR + "/MariaDB4j/base";
     protected String libDir = null;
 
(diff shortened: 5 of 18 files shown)


TheKnowles:
Related: https://github.com/vorburger/MariaDB4j/issues/344 https://github.com/vorburger/MariaDB4j/issues/699

DBs/pom.xml
```diff
@@ -20,16 +20,19 @@
     <packaging>pom</packaging>
 
     <modules>
+<!--        <module>mariaDB4j-db-win32-10.6.12</module>-->
```
vorburger, line 23: Why is this commented? You can (should) uncomment this, so that it can be tested by others - and eventually deployed.

DBs/pom.xml
```diff
@@ -20,16 +20,19 @@
...
         <module>mariaDB4j-db-win32-10.2.11</module>
         <module>mariaDB4j-db-win32-10.1.23</module>
         <module>mariaDB4j-db-win32-10.1.20</module>
         <module>mariaDB4j-db-win32-10.0.13</module>
 
-	<module>mariaDB4j-db-mac64-10.2.11</module>
+<!--        <module>mariaDB4j-db-mac64-10.6.12</module>-->
+	    <module>mariaDB4j-db-mac64-10.2.11</module>
```
vorburger, line 30: Clean this up and revert this effectively formatting only change here, to avoid confusion?

DBs/pom.xml
```diff
@@ -20,16 +20,19 @@
...
         <module>mariaDB4j-db-mac64-10.1.23</module>
         <module>mariaDB4j-db-mac64-10.1.9</module>
         <module>mariaDB4j-db-mac64-5.5.34</module>
 
+<!--        <module>mariaDB4j-db-linux64-10.6.12</module>-->
```
vorburger, line 35: Uncomment this (as above for the Windows version).

README.md
```diff
@@ -131,7 +131,7 @@ MariaDB4j JAR binaries are available from:
 <dependency>
     <groupId>ch.vorburger.mariaDB4j</groupId>
     <artifactId>mariaDB4j</artifactId>
-    <version>3.0.0</version>
+    <version>2.7.0</version>
```
vorburger, line 134: Nope, please revert this? 😄

README.md
```diff
@@ -146,7 +146,7 @@ source; just git clone this and then mvn install or deploy. -- MariaDB4j's Maven
 <dependency>
     <groupId>ch.vorburger.mariaDB4j</groupId>
     <artifactId>mariaDB4j</artifactId>
-    <version>3.0.0-SNAPSHOT</version>
+    <version>2.7.0-SNAPSHOT</version>
```
vorburger, line 149: revert please, as above

mariaDB4j-app/pom.xml
```diff
@@ -6,7 +6,7 @@
 	<parent>
 		<groupId>ch.vorburger.mariaDB4j</groupId>
 		<artifactId>mariaDB4j-pom</artifactId>
-		<version>3.0.0-SNAPSHOT</version>
+		<version>2.7.0-SNAPSHOT</version>
```
vorburger, line 9: nope, please revert

mariaDB4j-core/pom.xml
```diff
@@ -6,7 +6,7 @@
     <parent>
         <groupId>ch.vorburger.mariaDB4j</groupId>
         <artifactId>mariaDB4j-pom</artifactId>
-        <version>3.0.0-SNAPSHOT</version>
+        <version>2.7.0-SNAPSHOT</version>
```
vorburger, line 9: nope, please revert

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
```diff
@@ -109,7 +109,7 @@ protected ManagedProcess createDBInstallProcess() throws ManagedProcessException
         logger.info("Installing a new embedded database to: " + baseDir);
         File installDbCmdFile = configuration.getExecutable(Executable.InstallDB);
         ManagedProcessBuilder builder = new ManagedProcessBuilder(installDbCmdFile);
-        builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mysql_install_db"));
+        builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mariadb-install-db"));
```
vorburger, line 112: May I ask why?

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DBConfigurationBuilder.java
```diff
@@ -368,23 +368,23 @@ public DBConfigurationBuilder setExecutable(Executable executable, Supplier<File
     }
 
     protected Map<Executable, Supplier<File>> _getExecutables() {
-        executables.putIfAbsent(Server, () -> new File(baseDir, "bin/mysqld" + getExtension()));
-        executables.putIfAbsent(Client, () -> new File(baseDir, "bin/mysql" + getExtension()));
-        executables.putIfAbsent(Dump, () -> new File(baseDir, "bin/mysqldump" + getExtension()));
+        executables.putIfAbsent(Server, () -> new File(baseDir, "bin/mariadbd" + getExtension()));
+        executables.putIfAbsent(Client, () -> new File(baseDir, "bin/mariadb" + getExtension()));
+        executables.putIfAbsent(Dump, () -> new File(baseDir, "bin/mariadb-dump" + getExtension()));
```
vorburger, line 373: Oh, so the binaries all have new names in recent MariaDB versions? Interesting. But that then means that after merging this it will ONLY work with those new binaries. I'm torn about this, because the idea was that one could just put another binaries JAR on the classpath and it would still work. Hm... perhaps should be made configurable, then? Are you willing to work on this PR to introduce that? Or perhaps first raise a simpler separate PR proposing that?

mariaDB4j-maven-plugin/pom.xml
```diff
@@ -6,7 +6,7 @@
     <parent>
         <groupId>ch.vorburger.mariaDB4j</groupId>
         <artifactId>mariaDB4j-pom</artifactId>
-        <version>3.0.0-SNAPSHOT</version>
+        <version>2.7.0-SNAPSHOT</version>
```
vorburger, line 9: revert please, as above

mariaDB4j-springboot/pom.xml
```diff
@@ -6,7 +6,7 @@
     <parent>
         <groupId>ch.vorburger.mariaDB4j</groupId>
         <artifactId>mariaDB4j-pom</artifactId>
-        <version>3.0.0-SNAPSHOT</version>
+        <version>2.7.0-SNAPSHOT</version>
```
vorburger, line 9: revert please, as above

mariaDB4j/pom.xml
```diff
@@ -6,7 +6,7 @@
 	<parent>
 		<groupId>ch.vorburger.mariaDB4j</groupId>
 		<artifactId>mariaDB4j-pom</artifactId>
-		<version>3.0.0-SNAPSHOT</version>
+		<version>2.7.0-SNAPSHOT</version>
```
vorburger, line 9: revert please, as above

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/StartSimulatedForAllPlatformsTest.java
```diff
@@ -38,17 +38,21 @@
...
-    @Test public void simulatedStartOSX() throws Exception {
-        checkPlatformStart(DBConfigurationBuilder.OSX);
-    }
+    // Commented out for 10.6.12 binary addition
+    // With the decommission of bintray in 2021 and the move of OCI images to GitHub Content Repo
+    // as docker images, it was impossible to unpack a version for DBs/.
+    // See additional commentary in the PR.
+//    @Test public void simulatedStartOSX() throws Exception {
+//        checkPlatformStart(DBConfigurationBuilder.OSX);
+//    }
```
vorburger, line 55: Agreed, but isn't this actually unrelated to the purpose of this PR, to bump for Linux & Windows? Or do you "need" this, here?

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
```diff
@@ -174,7 +174,7 @@ protected String getReadyForConnectionsTag() {
 
     synchronized ManagedProcess startPreparation() throws ManagedProcessException, IOException {
         ManagedProcessBuilder builder = new ManagedProcessBuilder(configuration.getExecutable(Server));
-        builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mysqld"));
+        builder.setOutputStreamLogDispatcher(getOutputStreamLogDispatcher("mariadbd"));
```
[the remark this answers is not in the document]
TheKnowles: Starting with mariadb 10.5.2 the bootstrap script is mariadb-install-db.
Sym linking the older script mysql_install_db caused the unpacked jar to not have the correct platform sub directory.

mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
TheKnowles: Starting with mariadb 10.5.2 the bootstrap script is mariadb-install-db.
Sym linking the older script mysql_install_db caused the unpacked jar to not have the correct platform sub directory.

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleTutorialTest.java
```diff
@@ -86,15 +86,13 @@ protected void check(DBConfigurationBuilder config) throws SQLException, Managed
         db.start();
 
         String dbName = "mariaDB4jTest"; // or just "test"
-        if (!dbName.equals("test")) {
```
vorburger, line 89: Why this change? Do newer versions of MariaDB not "already have a DB named "test"? Then the inline comment below shouldn't be kept... This is actually kind of related to the other discussion, above; it will break anyone wanting to use older JARs. How about making this conditional of the MariaDB version?

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleTutorialTest.java
```diff
@@ -86,15 +86,13 @@ protected void check(DBConfigurationBuilder config) throws SQLException, Managed
...
-            db.createDB(dbName);
-        }
+        // mariadbd out-of-the-box already has a DB named "test"
+        // in case we need another DB, here's how to create it first
+        db.createDB(dbName);
 
         Connection conn = null;
         try {
-            conn = DriverManager.getConnection(db.getConfiguration().getURL(dbName), "root", "");
+            conn = DriverManager.getConnection(db.getConfiguration().getURL(dbName), System.getProperty("user.name"), "");
```
vorburger, line 95: This is interesting. I guess it's... cleaner? Would you be willing to submit a small PR just for this?

mariaDB4j/src/test/java/ch/vorburger/mariadb4j/tests/MariaDB4jSampleTutorialTest.java
```diff
@@ -130,15 +128,19 @@ public void testEmbeddedMariaDB4jWithSecurity() throws Exception {
...
-            // in case we need another DB, here's how to create it first
-            db.createDB(dbName, "root", "");
-        }
+        // mariadbd out-of-the-box already has a DB named "test"
+        // in case we need another DB, here's how to create it first
+        // Starting with MariaDB 10.4, the root user has an invalid password
+        // we will modify the root user password back to an empty string
+        // Using the user that owns the data directory / uid we can execute this initial bootstrapping command
+        // to give us pre 10.4 behavior for the purposes of this test
+        db.run("SET PASSWORD FOR 'root'@'localhost' = PASSWORD('root');", System.getProperty("user.name"), "");
```
vorburger, line 137: Oh, interesting. Because this test is also the "Tutorial", in a sense this promotes doing something "insecure". I'm actually a bit confused why the root user is still required if `System.getProperty("user.name")` is being use above. I wonder if there is a better solution for this. Could this entire TutorialTest just not use the (DB, not OS) `root` user at all? Perhaps improving this is worth a separate smaller PR, as a pre-requisite for this?

pom.xml
```diff
@@ -11,7 +11,7 @@
     </parent>
 
 	<artifactId>mariaDB4j-pom</artifactId>
-	<version>3.0.0-SNAPSHOT</version>
+	<version>2.7.0-SNAPSHOT</version>
```
vorburger, line 14: revert please, as above

vorburger (changes requested):
@TheKnowles Thank You for this!!! My apologies for the 3 week delay in reviewing. Unfortunately I can only work on this project occassionally. I'm hoping that my review feedback here is "motivating" to sort this out together to get it to a state where it can be merged, and not ... "harsh" in any way? I really do appreciate the contribution, but this PR will require some work. Also it needs to be rebased on the current `main` branch, as per related inline comment, and #717.

Issue #324, opened 2020-01-06
Remove /tmp/MariaDB4j at the end to enable easy multiple system users support


ghost:
If both temporary directories (`base` and `data`) are removed, maybe we could also remove the parent directory (`/tmp/MariaDB4j`). This would avoid the following issue:

1. run as `user1`
   - ok
2. run as `user2` (in my case, that's the `jenkins` user on a local Jenkins installation)
   - expected: no error
   - actual: `java.io.IOException: Unable to create directory /tmp/MariaDB4j/base`
     - cause: `/tmp/MariaDB4j` exists and it's only writable by `user1` (`rwxr-xr-x`)

vorburger:
Hello! 

Thank you for your interest in this project. 

You raise a fair point. If you would like to contribute to this project, please raise a Pull Request about this.

M.

vorburger:
FTR while (finally) updating the [CHANGELOG](https://github.com/vorburger/MariaDB4j/blob/main/CHANGELOG.md) today, I remembered that #604 added something related to this. That allows to "manually" do what this suggests - I guess the goal here would be for someone to "automate" it. Go for it and raise a PR!

vorburger:
@membersound @jzillmann @A248 thanks for your tumbs up (above) on this issue; FYI I'm cleaning up old issue, and as part of that closing this one which hasn't seen any community engagement in ~1 year; any future readers, please open a new issue, or (better) a PR with a fix!

PR #849, opened 2023-12-13
(#324) feat: hashcode tmp path

seunggabi:
#324 
divide tmp dir


vorburger:
@seunggabi Thank You for proposing this! Unfortunately it fails to build. Do you know how to check the logs and fix that?

I'll code review it more only once it builds; hope you understand and that's fair? Tx!

vorburger:
@seunggabi did you want to fix the broken build here?

Anyone else reading alone interested in cleaning this up?

It looks like a useful change that we could re-review and likely merge - if only it had a green build.

vorburger:
==> #985

PR #1043, opened 2024-10-02
Use execute_install_scripts: true

vorburger:


vorburger:
Does not help to fix problem; see #1045.

PR #1074, opened 2024-12-06
Run unit tests on ubuntu-24.04

ruroru:
libncurses5 issue was solved already

References: https://github.com/MariaDB4j/MariaDB4j/issues/1045

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -18,7 +18,7 @@ concurrency:
 jobs:
   testOnLinux:
     name: Maven Verify (Build & Test) on Linux
-    runs-on: ubuntu-22.04 # 24.04 is causing https://github.com/MariaDB4j/MariaDB4j/issues/1045
+    runs-on: ubuntu-24.04
     permissions:
       # maven-dependency-submission-action needs write
       contents: write


.github/workflows/test.yaml
```diff
@@ -18,7 +18,7 @@ concurrency:
 jobs:
   testOnLinux:
     name: Maven Verify (Build & Test) on Linux
-    runs-on: ubuntu-22.04 # 24.04 is causing https://github.com/MariaDB4j/MariaDB4j/issues/1045
+    runs-on: ubuntu-latest
```
vorburger, line 21: @ruroru Thank You very much for your contribution to this project! May I request that you use a fixed version here, so `ubuntu-24.04` instead of `ubuntu-latest`? The problem with using `-latest` versions, in general, not just for this project, is that it's "unstable" - when GitHub will change `ubuntu-latest` from say 24.04 to say 26.0 (or whatever) then the CI of the project could break again (without any changes having been made to the project) - which I don't want. I hope this makes sense? Thank You!

vorburger (changes requested):

vorburger:
> libncurses5 issue was solved already

Actually, the CI build fails, because of this...

Can you see (access) the log here?

Contributions to fix this very welcome, of course!

vorburger:
> libncurses5 issue was solved already

@ruroru on Dec 6, 2024 when you created this PR, this did not work, yet.

But today, after #1121 for #1120, this now works.

#1122 merged this; Thank You for your contrib!

PR #1096, opened 2025-02-02
Feature/support for conditional beans

luistrigueiros:


mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springframework/MariaDB4jSpringService.java
```diff
@@ -20,31 +20,20 @@
...
- * {@link Configuration}, yet let end-users override those via the Spring Values mariaDB4j.port,
- * mariaDB4j.socket, mariaDB4j.dataDir, mariaDB4j.baseDir; so e.g. via -D or (if using Spring Boot)
- * main() command line arguments.
- *
- * <p>This Service is intentionally NOT annotated as a {@link Service} {@link Component}, because we
- * don't want it to be auto-started by component scan without explicit declaration in a @Configuration
- * (or XML)
- *
- * @author Michael Vorburger
- */
```
vorburger, line 45: @luistrigueiros would it perhaps make sense to keep this JavaDoc? Or update it, instead of entirely delete it? You are welcome to add yourself as a (co) `@author` to it if you would, by the way.

vorburger:
@luistrigueiros Thank You for contributing to MariaDB4j! My apologies that I did not get a chance to review this before today.

I'm assuming that this is related to https://github.com/MariaDB4j/MariaDB4j/discussions/1084 ? Perhaps you would like to put an at least short description into PR, which I could use for the squashed commit, and in Release Notes, to document this change?

This would also need to be green on CI, and pass all the tests. Do you have access to the build log failures, and would be willing to look into getting it to pass? The problem on the failed Linux build job check is something which I believe I have meanwhile fixed, so if I could rebase this PR please, that should just go away. (Feel free to ask me here if you need any support about how to rebase in git.) The problem on the Windows build job check with all those JavaDoc failures I don't quite understand (because you haven't changed that here?), but again if you rebase it's possible that I've fixed it meanwhile - it's worth a try.

mariaDB4j-springboot/src/main/java/ch/vorburger/mariadb4j/springframework/MariaDB4jSpringService.java
luistrigueiros: Thank @vorburger I will issue a PR to update this

luistrigueiros:
I confess to have found some difficulties  with updating the JavaDocs and having then passing in my machine not breaking the build as there looks to be something that is performing checks in the the  comments  l

vorburger:
> I confess to have found some difficulties with updating the JavaDocs and having then passing in my machine not breaking the build as there looks to be something that is performing checks in the the comments

I had a quick closer look for you: It may be a bit confusing because there are so many messages? But you can IGNORE all of those "warning: no comment" - those were there before, and you don't have to fix them as part of this. The only thing that makes this fail the build is:

```
/home/runner/work/MariaDB4j/MariaDB4j/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/MariaDB4jService.java:40: error: reference not found
Error:   * @see MariaDB4jSpringService
Error:          ^
```

If you fix that, I guess you'll want to just remove it (?), it should hopefully look much better! I hope this helps to unblock you.

luistrigueiros:
Thank @vorburger fixed it now

luistrigueiros:
Thank you @vorburger

Issue #1107, opened 2025-03-03
Broken Build


vorburger:
The CI build is completely hosed.

It has something to do with https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

But I couldn't immediately see where we're still using an old version of `actions/checkout@v3`.

[Everywhere I looked](https://github.com/search?q=repo%3AMariaDB4j%2FMariaDB4j%20actions%2Fcheckout%40&type=code) we're on (latest) `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2` already.

Can anyone else reading along here and wanting to help maintain this project spot what the problem is?

vorburger:
> It has something to do with https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

> Getting action download info
> Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

vorburger:
> [Everywhere I looked](https://github.com/search?q=repo%3AMariaDB4j%2FMariaDB4j%20actions%2Fcheckout%40&type=code) we're on (latest) actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2 already.

Wait, I got confused, this is not about `actions/checkout`, but [`download-artifact`](https://github.com/search?q=repo%3AMariaDB4j%2FMariaDB4j+actions%2Fdownload-artifact%40&type=code) (not used) and [`actions/upload-artifact`](https://github.com/search?q=repo%3AMariaDB4j%2FMariaDB4j+actions%2Fupload-artifact%40&type=code) - at v4.4.3, when #1105 proposes a bump to #4.6.1. 

I could just merge that, but... it's in the wrong action, we only use it in Scorecard, but what fails is the main build job.

This is strange.

vorburger:
Oh... wait, does this:

```
Current runner version: '2.322.0'
Operating System
Runner Image
Runner Image Provisioner
GITHUB_TOKEN Permissions
Secret source: Dependabot
Prepare workflow directory
Prepare all required actions
Getting action download info
[...]

PR #1108, opened 2025-03-04
Remove step-security/harden-runner (re. #1107)

vorburger:


vorburger:
The cause of #1107 is not `step-security/harden-runner` but `awalsh128/cache-apt-pkgs-action` - see #1109.

PR #1109, opened 2025-03-04
Remove awalsh128/cache-apt-pkgs-action

vorburger:
See https://github.com/MariaDB4j/MariaDB4j/issues/1107.


vorburger:
Not needed, see https://github.com/MariaDB4j/MariaDB4j/issues/1107#issuecomment-2696216104.

PR #1112, opened 2025-03-28
dev: DevContainer with Java 17 (instead of 11) 

vorburger:
To be able to [more easily] use (e.g.) GitHub Codespaces.


vorburger:
FTR: This fixes Java 17 vs. Java 11, but tests still fail, due to the `libncurses.so.5` thing; but that's separate.

PR #1085, opened 2025-01-25
Homebrew support

abrousseau001:
Hopefully this can be a starting point for #497. There is still work to do, but I can't currently test it on my mac because there is a strange issue with running the tests.

The issue appears similar to this https://github.com/spring-projects/spring-boot/issues/6421

PR #1123, opened 2025-03-28
Homebrew support

vorburger:
#1085 carry-over, with some additions. Still ToDo:

- [ ] Test this actually works on MacOS, using https://flow.swiss/mac-bare-metal
- [ ] Launch `brew` itself using [ch.vorburger.exec](https://github.com/vorburger/ch.vorburger.exec) (just like MariaDB4j launches `mariadbd`) [for better logging, etc.]
- [ ] Properly re-review code, and likely push a few more amendments (e.g. do not "double log & throw")
- [ ] Add CI coverage with GitHub Mac OS Runner
- [ ] Document it in the README

### DIFF
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
@@ -56,6 +56,8 @@ public class DB {
 
     private static final Logger logger = LoggerFactory.getLogger(DB.class);
 
+    private static final String homebrewInstallationPath = "/opt/homebrew/bin/brew";
+
     protected final DBConfiguration configuration;
 
     private File baseDir;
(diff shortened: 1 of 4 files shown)


abrousseau001:
I am glad you are taking a look at this @vorburger. I wish I could have gotten it to build locally or figure out what the issue was, it shouldn't have been modified too much.

My daughter has my MacBook Pro now after her Chromebook died so until I get a replacement, my macOS testing days are on hold.

I still think that we should consider whether this is the right option since by having homebrew as a requirement now means effectively admin access to at least install homebrew that first time. I honestly just wish MariaDB was natively built for macOS as well, but that doesn't look like it will ever happen.

vorburger:
> My daughter has my MacBook Pro now after her Chromebook died so until I get a replacement, my macOS testing days are on hold.

And she would not lend it to you, anymore? 🤣  [Mine is the same!] 👸🏽 

https://flow.swiss/mac-bare-metal works great, at least for me; BTW. (I've just set it up; not run this, yet.)

> I still think that we should consider whether this is the right option since by having homebrew as a requirement now means effectively admin access to at least install homebrew that first time. I honestly just wish MariaDB was natively built for macOS as well, but that doesn't look like it will ever happen.

Wehl... "it's complicated", isn't it? While https://mariadb.org/download/ do not distribute MariaDB executable binaries for macOS, other community do offer native builds, such as (the very awesome!) [Nix-ians](https://github.com/NixOS/nixpkgs/tree/master/pkgs/servers/sql/mariadb), and Homebrew - as discussed in #497.

If this works as-is, I'm actually not against merging this already - it's better than what there is today? In your view, should we not?

Alternatively, or in addition, I may try some of the tips provided in #497 as well.

vorburger:
> If this works as-is

As is, for me it fails, due to:

```
java.io.IOException: Cannot run program "/opt/homebrew/bin/brew list mariadb": error=2, No such file or directory
        at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1143)
```

BTW if we pursue this, then we should launch `brew` itself using [ch.vorburger.exec](https://github.com/vorburger/ch.vorburger.exec) (just like MariaDB4j launches `mariadbd`) - for better logging, etc.

> by having homebrew as a requirement now means effectively admin access to at least install homebrew that first time

That's actually an interesting great point! I'm not super familiar with the UX on macOS, but yeah if on Linux as a Java user some "library" (like MariaDB4j) would expect `root`, then that would ... not be cool, at all (at least for me). @Muta-Jonathan what's your view on this, if any?

> others do offer native builds, such as ... Homebrew - as discussed in #497.

I'm going to "hold" on this PR as-is (at least for the moment), and explore that idea / approach a bit more now...

vorburger:
@abrousseau001 I'm closing this (unmerged), based on the "break through" in #1124 - hope OK for you.

abrousseau001:
I am happy with your solution instead. It feels like a better fit with the spirit of the project.

Just like we discussed before, having homebrew installed in constrained environments like I worked before, makes this harder on macOS. Maybe one day my friend @vorburger.

Muta-Jonathan:
thank you @vorburger for putting this as a priority and @abrousseau001 for the initial work

PR #1124, opened 2025-03-28
db: Add mariaDB4j-db-11.4.5-macos-arm64 (see #497)

vorburger:
See #497!

🍏 👏🏽

PR #1125, opened 2025-03-29
ci: Add macOS CI test coverage by adding macos-15 GitHub Action Runner

vorburger:


vorburger:
I don't understand why this PR did not run the `.github/workflows/test.yaml` at all anymore?! 🤯 

Is that a security pre-caution by GitHub? (When you change CI, it doesn't run on PR, only after merge?)

vorburger:
Even after (separately, first; only) merging #1124, and this now only having the `.github/workflows/test.yaml` it still does not build...

... because I forgot a missing `steps:` in the YAML! Duh.

PR #1127, opened 2025-03-29
ci: Add action-validator pre-commit

vorburger:


vorburger:
Ah, this breaks https://results.pre-commit.ci/run/github/3355640/1743209397.nQiTXlOAQmyV5Y1uxbZEqQ :

```
Detect hardcoded secrets.................................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
action validator.........................................................Failed
- hook id: action-validator
- exit code: 1

action-validator not installed or available in the PATH
```

Well, so in that case, no then; it's "not worth it".

PR #1134, opened 2025-03-29
db: Package only mysql*.exe but not mariadb*.exe on winx64

vorburger:
Follow-up to https://github.com/MariaDB4j/MariaDB4j/pull/1126/files#r2019771660,

to save ~17 MB (~27%) on JAR file size.


vorburger:
@TheKnowles FYI & check this out! 🗜️

PR #1126, opened 2025-03-29
Windows MariaDB 11.4.5 upgrade.

TheKnowles:
Adds Windows MariaDB 11.4.5 binaries.

PR #1136, opened 2025-03-29
ci: Do not mvn -f DB anymore

vorburger:
Now that I've deployed `winx64` from #1126 (with #1134) to Maven Central, this is no longer required (and would only be confusing, in the future).


vorburger:
@TheKnowles FYI and let's do future upgrades using this approach as well, I think this is a good "process" how we can go about it.

PR #1156, opened 2025-04-09
(Fix) AssertionError: expected:<5678> but was:<49961> at ch.vorburger.mariadb4j.tests.springframework.*

Muta-Jonathan:
Fixes https://github.com/MariaDB4j/MariaDB4j/issues/1150

The problem 
---
was in the MariaDB4jSpringService class. The setDefaultPort method 
was not being picked up by the `@Value("${" + MariaDB4jSpringService.PORT + ":-1}")` injection because `@Value` was evaluated before the test-specific configuration is applied. That’s why the builder falls back to a random open port instead of the configured one.

### DIFF
--- a/mariaDB4j-springboot/pom.xml
+++ b/mariaDB4j-springboot/pom.xml
@@ -37,4 +37,21 @@
         <scope>test</scope>
     </dependency>
   </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>3.5.3</version>
+        <configuration>
+          <includes>
+            <include>**/*Test.java</include>
+          </includes>
+          <excludes>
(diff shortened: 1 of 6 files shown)


mariaDB4j-springboot/pom.xml
```diff
@@ -37,4 +37,21 @@
         <scope>test</scope>
     </dependency>
   </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>3.5.3</version>
```
vorburger, line 45: It would be nice if this version here could be omitted instead of duplicated from the parent, if possible. 

But it's too minor to hold up merging this, I can follow up with a separate PR later.

vorburger (approved):

vorburger:
@Muta-Jonathan this is AWESOME!! Thank You so much.

Issue #1042, opened 2024-10-01
Reusing data dir


xtianus:
I'd like to reuse the data dir on multiple runs in order to keep my data.

Currently this can't be done because the `DB.newEmbeddedDB()` method always tries to initialize a new db in the data folder, which must be empty or I get the error

    ch.vorburger.exec.ManagedProcess - : mariadb-install-db.exe: ERROR : Data directory C:\somefolder\mydata is not empty. 
    Only new or empty existing directories are accepted for --datadir

So I forked the repo and implemented a `openEmbeddedDB()` method 

    public static DB openEmbeddedDB(DBConfiguration config) throws ManagedProcessException {
        DB db = new DB(config);
        db.prepareDirectories();
        return db;
    }

but I still can't run the database on the same folder twice if I stop my java process with CTRL-C (on windows) because the mariadb.exe process is not stopped by the shutdown hooks and keeps running, preventing a restart of my application with the error

    InnoDB: The data file './ibdata1' must be writable

as the mariadb.exe process locks ibdata1.

So I'm wondering if there's a way to reattach to an existing mariadb process that has been previously started. This may be a feature request I guess.

vorburger:
@xtianus Thank You for your interest in this library, and providing this feedback. Would you be interested to contribute a PR which does the needful for you?

vorburger:
#1042 sounds related BTW. Contributions to address this are welcome.

xtianus:
I'd be happy to implement a fix but it looks like there's nothing we can do about it.

vorburger:
@xtianus @Muta-Jonathan I'm assuming that [your!] the (just) merged #1166 addressed this; please comment to re-open if you disagree.

PR #1166, opened 2025-04-16
Open existing database

xtianus:
Issue #1042
Added a DB.openEmbeddedDB() method to reuse an existing database.

The use case would be:
- check if the database folder exists
- when it does not exist, use the DB.newEmbeddedDB() method
- when it exists, use the DB.openEmbeddedDB() method

This will ensure that you can restart your process without losing stored data.


vorburger (changes requested):
@xtianus Thank You for contributing to this project! This change is interesting, but it's "important (big) enough" that I cannot accept it without any test coverage; would you be wiling to add a new `@Test` illustrating this new functionality in the existing `MariaDB4jSampleTutorialTest`, please?

Also, could you remove the 4 to 2 space indention change again?

vorburger:
@xtianus did you want to do this? (Just checking if you are still interested in implementing the feedback.)

Muta-Jonathan:
@vorburger any update on this Am getting on only windows

> [INFO] mysql_install_db.exe: Running bootstrap
> [ERROR] mysql_install_db.exe: ERROR : Data directory F:\Projects\openmrs-standalone\target\emptydatabase\data is not empty. Only new or empty existing directories are accepted for --datadir
> [ERROR] Program [F:\Projects\openmrs-standalone\target\emptydatabase\bin\mysql_install_db.exe, --datadir=F:\Projects\openmrs-standalone\target\emptydatabase\data] (in working directory F:\Projects\openmrs-standalone\target\emptydatabase) failed unexpectedly
> org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
>     at org.apache.commons.exec.DefaultExecutor.executeInternal (DefaultExecutor.java:355)
>     at org.apache.commons.exec.DefaultExecutor.lambda$execute$0 (DefaultExecutor.java:269)
>     at java.lang.Thread.run (Thread.java:842)
> [ERROR] Program [F:\Projects\openmrs-standalone\target\emptydatabase\bin\mysql_install_db.exe, --datadir=F:\Projects\openmrs-standalone\target\emptydatabase\data] (in working directory F:\Projects\openmrs-standalone\target\emptydatabase) failed
 Is this able to solve the issue on windows @xtianus

vorburger:
@Muta-Jonathan it looks like @xtianus lost interest in this, and hasn't been able to update it to take the code into account.

But if this of interest & value to you, then what you should do is to create a new PR from you, based on this, with the missing test, and intendtation undone! Then I would review that one from you instead.

vorburger:
> this method deletes the db when it is in a temp folder:

How about using a relative directory of (something like) `target/startAndStopThenReopenExistingTest/db-bin | data`? As far as I know, Surefire sets the current working directory to the module root (where the `pom.xml` is), and the `target/` directory seems like a good destination for this, no? `isTemporaryDirectory` won't consider that to be _"temp"._ Try it, if you think that could work?

Muta-Jonathan:
@xtianus @vorburger this is a blocker on our end could we priotise fixing this?

vorburger (approved):

PR #1172, opened 2025-04-23
build: Run google-java-formatter as pre-commit

vorburger:


vorburger:
Closing, since the use of the `fmt-maven-plugin`, integrated into the Maven build, meanwhile provides similar (enough) functionality.

Issue #926, opened 2024-04-04
3.1.0 - "<path>" is not empty. Only new or empty existing directories are accepted for --datadir


Osiris-Team:
```java
    public static void initIntegratedMariaDB() throws ManagedProcessException {
        DBConfigurationBuilder configBuilder = DBConfigurationBuilder.newBuilder();
        configBuilder.setPort(0); // OR, default: setPort(0); => autom. detect free port
        configBuilder.setDataDir(new File(System.getProperty("user.dir") + "/db").getAbsolutePath());
        db = DB.newEmbeddedDB(configBuilder.build());
        db.start();
        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
            try {
                System.out.println("Stopping DB...");
                db.stop();
                System.out.println("Stopped DB successfully.");
            } catch (ManagedProcessException e) {
                throw new RuntimeException(e);
            }
        }));
    }
```
```sh
2024-04-04 09:20:10.521 [main] INFO ch.vorburger.mariadb4j.Util - Created directory: C:\Users\arman\AppData\Local\Temp\MariaDB4j\base
2024-04-04 09:20:10.526 [main] INFO ch.vorburger.mariadb4j.Util - Created directory: C:\Users\arman\AppData\Local\Temp\MariaDB4j\base\libs
2024-04-04 09:20:10.527 [main] INFO ch.vorburger.mariadb4j.Util - Created directory: C:\Users\arman\AppData\Local\Temp\MariaDB4j\tmp\51248
2024-04-04 09:20:12.052 [main] INFO ch.vorburger.mariadb4j.Util - Unpacked 80 files from classpath*:ch/vorburger/mariadb4j/mariadb-10.11.5-fix1/winx64/** to C:\Users\arman\AppData\Local\Temp\MariaDB4j\base
2024-04-04 09:20:12.052 [main] INFO ch.vorburger.mariadb4j.DB - Installing a new embedded database to: C:\Users\arman\AppData\Local\Temp\MariaDB4j\base
2024-04-04 09:20:12.097 [main] INFO ch.vorburger.exec.ManagedProcess - Starting Program [C:\Users\arman\AppData\Local\Temp\MariaDB4j\base\bin\mariadb-install-db.exe, --datadir=C:\Users\arman\IdeaProjects\daniel_manager\db] (in working directory C:\Users\arman\AppData\Local\Temp\MariaDB4j\base)
[...]

PR #1217, opened 2025-08-03
clean: Remove openEmbeddedDB(int port) [but keep DBConfiguration variant]

vorburger:
Remove again the overloaded 2nd method that was (just) introduced in #1166 for #1042.

The `openEmbeddedDB(DBConfiguration config)` variant makes sense - and is used in the `testEmbeddedMariaDB4jReopenExisting()`.

But I'm not sure that I understand what/when/how the `openEmbeddedDB(int port)` API would ever be used for... especially since #1145, isn't that... (almost) "impossible" to be useful in reality? As in, the directory paths would always be different anyways... unless you re-used literally the exact same identical Java object, which... isn't that kind of pointless then? (My understanding is that the interest of this in practice is to use it ACROSS different JVM starts, with paths that have been fixed by other means of configuration.)

I thought about this more only after merging #1166 (and didn't want to hold that long overdue PR up just because of this).

@xtianus @Muta-Jonathan @Osiris-Team @TheKnowles @wuxindao do any of you have any objection to this PR which (already) removes the `openEmbeddedDB(int port)` API variant again? (We would **_keep_** the other `DB openEmbeddedDB(DBConfiguration config)` variant - obviously.)

### DIFF
--- a/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
+++ b/mariaDB4j-core/src/main/java/ch/vorburger/mariadb4j/DB.java
@@ -2,7 +2,7 @@
  * #%L
  * MariaDB4j
  * %%
- * Copyright (C) 2012 - 2017 Michael Vorburger
+ * Copyright (C) 2012 - 2025 Michael Vorburger
  * %%
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -93,21 +93,6 @@ public class DB {
         DB db = new DB(config);
         db.prepareDirectories();
         return db;
-    }
-
-    /**
-     * This factory method is the mechanism for opening an existing embedded database for use. This
-     * method assumes that the database has already been prepared for use with default
-     * configuration, allowing only for specifying port.
-     *
-     * @param port the port to start the embedded database on
-     * @return a new DB instance
-     * @throws ManagedProcessException if something fatal went wrong
-     */
-    public static DB openEmbeddedDB(int port) throws ManagedProcessException {
-        DBConfigurationBuilder config = DBConfigurationBuilder.newBuilder();
-        config.setPort(port);
-        return openEmbeddedDB(config.build());
     }
 
     /**


Muta-Jonathan:
Basically on our end (Openmrs) this is not used just as you expressed I also see no need of it to be used ... to me this makes sense 

cc @dkayiwa @vorburger

Osiris-Team:
@vorburger Yeah im not sure why that open method was introduced in the first place since the configuration already has an option
to avoid deletion and specify directories to reuse existing data, at least in mariadb4j 3.0.1, which is the version I was forced to stay on since after that it was impossible to open existing databases #926.

Here is how I use it:
```java
/**
 * Creates or uses the database inside ./db and runs it via MariaDB4j on a random available port. <br>
 * MariaDB4j handles downloading of MariaDB and launching it. <br>
 * Returns once fully launched or throws exception on fail. <br>
 */
public static void initIntegratedMariaDB() {
    try{
        DBConfigurationBuilder configBuilder = DBConfigurationBuilder.newBuilder();
        configBuilder.setPort(0); // OR, default: setPort(0); => autom. detect free port
        configBuilder.setDataDir(new File(System.getProperty("user.dir") + "/db").getAbsolutePath());
        configBuilder.setDeletingTemporaryBaseAndDataDirsOnShutdown(false);
        mariaDB = DB.newEmbeddedDB(configBuilder.build());
        mariaDB.start();
        String port = url.substring(url.lastIndexOf(":"), url.lastIndexOf("/"));
        url = url.replace(port, ":"+mariaDB.getConfiguration().getPort());
        rawUrl = getRawDbUrlFrom(url);
        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
            try {
                System.out.println("Stopping database...");
                mariaDB.stop();
                System.out.println("Stopped database successfully.");
            } catch (ManagedProcessException e) {
                throw new RuntimeException(e);
            }
        }));
    } catch (Exception e) {
        throw new RuntimeException(e);
    }
}
```

Osiris-Team:
Instead of keeping/exposing the other `openEmbeddedDB(DBConfiguration config)` you could also simply check if a data dir was provided and deletion was disabled, which means that the database should be persistent and then call that method internally instead.

Muta-Jonathan (approved):
to me looks fine

PR #1401, opened 2026-09-08
Add option rootDir

joca-bt:
It's difficult to change the root dir for the mariadb4j files. If the user only wants to change the root dir while keeping the default layout, this requires changing several options and mirroring what the library actual does.

Introduce option rootDir to simplify this.

Options baseDir, libDir, dataDir, tmpDir, and socket all had hardcoded paths containing java.io.tmpdir, this is now less problematic.


joca-bt:
hello @vorburger, could you have a look at this proposal?

